Issue #1, opened 2021-01-18
Update Api


Freymaurer:
To update the current API state we need to bring some of the type restrictions up to date with the DataModel changes done in previous commits. In addition we should redistribute the API modules over the .fs files in the API folder.

- [x] Update type restrictions.
- [x] Redistribute API modules over .fs files (leave API.fs for now as deprecated)
- [x] As discussed i added a general Record Type update type (See Update.fs) This allows a user to define how a record type should be updated.
- [x] Add unit tests for Update.fs.

PR #2, opened 2021-01-18
Update Api (Issue #1).

Freymaurer:
This PR updates the API module according to Issue #1.

Please let me know if i should change anything.


tests/ISADotnet.Tests.NetCore/ISADotnet/API/UpdateTests.fs
```diff
@@ -0,0 +1,252 @@
+﻿module UpdateTests
```
HLWeil, line 1: Module name APITests would be more appropriate

src/ISADotnet/InvestigationFile/API.fs
```diff
@@ -1,640 +1,634 @@
 ﻿namespace ISADotNet.InvestigationFile
```
HLWeil, line 1: The functions need to be resorted. It does not get clear from the file and module structur which function is related to studies and which is related to investigations

src/ISADotnet/API/Update.fs
```diff
@@ -0,0 +1,117 @@
+﻿module Update
```
HLWeil, line 1: Could you add this to a private module within namespace ISADotNet?

Also could you add some comments?

HLWeil (changes requested):

tests/ISADotnet.Tests.NetCore/ISADotnet/API/UpdateTests.fs
Freymaurer: The unit tests i added are only to test the functionality written in Update.fs. If you want to add API tests to this .fs in the future i will of course change the name.

src/ISADotnet/InvestigationFile/API.fs
Freymaurer: This file is only left in as depricated  to get the previously done changes into the commit history. I did this to have a failback safe in case i missed to redistribute something or did an error during the merge of your newest commits. So if something went wrong we have this file with some of the updates still in the commit history. This is also the reason why i added the obsolete tag to the API module. I hope this gives some additional insight. Let me know what you think.

tests/ISADotnet.Tests.NetCore/ISADotnet/API/UpdateTests.fs
HLWeil: Yes, other API tests will surely follow and I think it might be more comprehensible to group them in a APITests module

HLWeil (approved):

HLWeil:
Seems good. I will include some final clean up steps when preparing the project for packaging

Issue #4, opened 2021-01-24
Update function not working for record in record


HLWeil:
# Problem
When updating one record type with another using the methods in `API.Update`, even when using the `UpdateByExisting` option, a field with a filled record is replaced by a record where no value is set.

An example for this is the `Measurementtype` field of the `Assay` type

# Solution

There should be a way to detect empty records and ignore them. Possibly using an `option` for records in records

Freymaurer:
This is actually by design, as i found it to exhausting to check for all types inside the record type again and then only update it if it contains a new value. So record types will always be replaced. 

> Possibly using an option for records in records

If you define the field as `Measurementtype option` this will work already.

HLWeil:
Yes I thought so, and that's fine. Checking the record recursively kind of imposes a new batch of assumptions, which might not be quite transparent when applying the function.

Nice to hear though that the `option` case is already covered.

PR #7, opened 2021-02-03
Append optional list when updating record types

HLWeil:
Some minor adjustemts were made to the `record type update` functionality. When set to `UpdateAllAppendLists`, also `optional lists` will be `appended`. This became necessary after all fields in the datamodel were set to optional.
Additonally, tests were added to cover these cases.


HLWeil:
closes #4

Issue #15, opened 2021-02-12
[Feature Request] Assay file reader


HLWeil:
Assay files are modified using the swate tool. To access this information for other tasks, a reader should be added. 
As the swate tool adds tables to assay xlsx files, a prerequisite for this reader is a table reader in FSharpSpreadsheetML. The information could be stored as grouped processes.

HLWeil:
Additionally, the assay file can store other metadata in a specific sheet. This way the assay is more self contained.

PR #17, opened 2021-03-07
Assay file io

HLWeil:
Added an ISAXLSX assay file reader, which is able to parse assay tables created with the [Swate](https://github.com/nfdi4plants/Swate) tool.


HLWeil:
closes #15

Issue #34, opened 2021-09-10
Worksheet name parsing from assay to CommonAPI rowMajor mistake


Freymaurer:
I translated one of the existing Swate templates to the new preview format and used the new common api row major format with it.
In Assay the worksheet name was correct, but after parsing to RowWiseSheet it was wrong as `1SPL01_plants` was changed to `1SPL01`.

The problem lies in https://github.com/nfdi4plants/ISADotNet/blob/AssayFileIO/src/ISADotnet/JsonIO/AssayCommonAPI.fs#L98 .

I am currently working on two different versions of a fix.

HLWeil:
Hey, good catch. That's a really unnecessary mistake on my side.
Grouping should just be done by the protocol name. 
Something like:
```fsharp
|> List.groupBy (fun x -> x.ExecutesProtocol. Value.Name. Value)
```

Issue #35, opened 2021-09-10
Term Accession of most parsed column headers is done incorrectly.


Freymaurer:
Currently when parsing s Swate AnnotationTable with ISADotNet to any ISA-JSON format. Ontology terms will be parsed as the following. Term Accession should be in uri format and can be created by combining the existing TSR and TAN with a static obo purl url: 
```json
"category": {
      "characteristicType": {
          "annotationValue": "Sample type",
          "termSource": "NFDI4PSO",
          "termAccession": "0000064"
      }
  }
```

Purl URL:

`"http://purl.obolibrary.org/obo/"`

Currently working on a PR to add this.

PR #36, opened 2021-09-10
Fix issues #34 and #35

Freymaurer:
#34, #35, #37

- [x] Tests

### DIFF
--- a/src/ISADotNet.XLSX/CollectionAux.fs
+++ b/src/ISADotNet.XLSX/CollectionAux.fs
@@ -104,7 +104,11 @@ module internal List =
 
 module internal Dictionary =
 
-    let tryGetValue k (dict:System.Collections.Generic.Dictionary<'K,'V>) = 
+    let tryGetValue k (dict:System.Collections.Generic.Dictionary<'K,string>) = 
         let b,v = dict.TryGetValue(k)
-        if b then Some v
-        else None
+        // Only get value if 
+        if b && v.Trim() <> ""
+        then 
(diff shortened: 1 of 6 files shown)


HLWeil:
Hey, thanks alot for the fixes! 🚀 Maybe you could include #37 too?

Freymaurer:
Done!

src/ISADotNet.XLSX/CollectionAux.fs
```diff
@@ -104,7 +104,11 @@ module internal List =
 
 module internal Dictionary =
 
-    let tryGetValue k (dict:System.Collections.Generic.Dictionary<'K,'V>) = 
+    let tryGetValue k (dict:System.Collections.Generic.Dictionary<'K,string>) = 
         let b,v = dict.TryGetValue(k)
-        if b then Some v
-        else None
+        // Only get value if 
+        if b && v.Trim() <> ""
```
HLWeil, line 110: If I see it correctly this will only trim the string it does not only contain whitespace, right? I think a string containing only whitespace should be trimmed too?

For this the second condition `(v.Trim() <> "")` should be removed

HLWeil (changes requested):

HLWeil (approved):

src/ISADotNet.XLSX/CollectionAux.fs
HLWeil: Nvm the `None` does the job

PR #58, opened 2022-04-05
Bugfixes and minor improvements

### LINKED ISSUE #50: [BUG] ProcessSequence.getOutputsWithCharacteristicBy does not return correct Characteristics
**Describe the bug**
The function ISADotNet.API.ProcessSequence.getOutputsWithCharacteristicBy does not return the Characteristics and Outputs of the given protocol. Instead, the Characteristics of the Protocol where the outputs serve as Input are returned.

**To Reproduce**
- Take an isa assay file with at least two protocols and one characteristic each
- Try to retrieve the characteristic of protocol one with the function ISADotNet.API.ProcessSequence.getOutputsWithCharacteristicBy

**Expected behavior**
The function should return the chraracteristics of the protocol that match predicate.

### LINKED ISSUE #52: [BUG] Unexpected ISADotNet.Viz Assay representation
**Describe the bug**
This is a follow up issue to #51. I added an input column `Source Name` and put the previous information from `Data File Name` in there and added artificial names to the now empty output column. These columns now look like this:

| Data File Name | Source Name                                |
|----------------|--------------------------------------------|
| result1        | DB_097_CAMMD_CAGATC_L001_R1_001.fastq.gz   |
| result2        | DB_099_CAMMD_CTTGTA_L001_R1_001.fastq.gz   |
| result3        | DB_103_CAMMD_AGTCAA_L001_R1_001.fastq.gz   |
| result4        | DB_161_reC3MD_GTCCGC_L001_R1_001.fastq.gz  |
| result5        | DB_163_reC3MD_GTGAAA_L001_R1_001.fastq.gz  |
| result6        | DB_165_re-C3MD_GTGAAA_L002_R1_001.fastq.gz |

The DAG will now be displayed, but i found two odd occurences:

image

As you can see the DAG shows that sheet 3 and 4 are applied twice. I cannot find a reason why this should be inteded, maybe you can help me out here.

### LINKED ISSUE #54: [BUG] Assay from Investigation File and Assay from AssayFile are not the same even if they are the same
**Describe the bug**
When using `AssayFile.Assay.fromFile` and `(Investigation.fromFile <path>).Studies.Value.Assays.Value.[i].Value`, the resulting assays are not identical even if they were initialized together in one ARC (e.g. via ArcCommander).

**To Reproduce**
1. Initiate ARC
2. Create Investigation
3. Add Assay
4. Use `arc a edit -a <assayID>`

**Expected behavior**
Assay objects are identical.

**Screenshots**
image

### LINKED ISSUE #56: [BUG] Assay: Headers with strange characters are not parsable
**Describe the bug**
If an Assay consists of headers (regardless whether Parameter, Factor, or Characteristic) which contains unusual characters (namely `(` or `[`) the column cannot be parsed.

**To Reproduce**
1. Create Assay with column headers `hello (world)` and `hello [world]`
2. Try parse with common ISA.NET functions
3. See error

**Expected behavior**
Either
- parse correctly, or
- throw error or warning that the respective column cannot be parsed

HLWeil:
### Assay File Header parsing

Removed some failing cases of header parsing (e.g. when there were brackets inside the name). Examples for these new rules are seen below:

image

`#` are now also included in the name together with potential numberings. Explanation see next section.

### Dropping number handling

`ISADotNet` should follow more closely the `ISATab` specification and not impose any unncessary additional assumptions when simply parsing assay files. Therefore from now on ISADotNet will move towards interpreting headers with `Swate` style numbering (e.g. `Parameter [temperature unit#2]`) oblivious to these numberings. 

Members incoporating these numbering rules (e.g. `Factor.GetNameAsStringWithNumber`) are now marked as `deprecated` and will be omitted in future releases.

### XLSX File IO verbosity
Failing xlsx readers will now be more verbose about why they are failing

### Small fixes

- dag vizualization now makes edges distinct per source target pair
- include empty study by default when writing investigation file
- fix assay reading no processes as Some empty lists
- fix outputByCharacteristic getter function

-----------------
Header parsing and number handling changes might be of interest to you:
@omaus, @Freymaurer, @Joott, @muehlhaus 

----------------

fixes #50
fixes #52
fixes #54 
fixes #56

### DIFF
--- a/src/ISADotNet.Viz/DAG.fs
+++ b/src/ISADotNet.Viz/DAG.fs
@@ -25,6 +25,7 @@ type DAG =
                 p.Outputs.Value
                 |> List.zip p.Inputs.Value
                 |> List.map (fun (i,o) -> ProcessInput.getName i ,ProcessOutput.getName o ,p.Name.Value)
+                |> List.distinct
             )
             |> List.append (rootNodes |> List.map (fun i -> "Root",i,""))
                           
--- a/src/ISADotNet.XLSX/AssayFile/AnnotationColumn.fs
+++ b/src/ISADotNet.XLSX/AssayFile/AnnotationColumn.fs
@@ -31,8 +31,8 @@ module AnnotationColumn =
         /// Parses a string to a column header
         static member fromStringHeader header =
                   
-            let kindPattern = @".*(?= [\[\(])"
-            let namePattern = @"(?<= \[)[^#\]]*(?=[\]#])"
+            let kindPattern = @"[^\[(]*(?= [\[\(])"
+            let namePattern = @"(?<= \[).*(?=[\]])"
             let ontologySourcePattern = @"(?<=\()\S+:[^;)#]*(?=[\)\#])"
             let numberPattern = @"(?<=#)\d+(?=[\)\]])"
 
--- a/src/ISADotNet.XLSX/AssayFile/AnnotationColumn.fs
+++ b/src/ISADotNet.XLSX/AssayFile/AnnotationColumn.fs
@@ -10,6 +10,16 @@ module AnnotationColumn =
     [<Literal>]
     let OboPurlURL = @"http://purl.obolibrary.org/obo/"
 
+    module RegexPattern =
+
+        // Pattern matches start of line and then any number of characters except newline, open bracket and open curly bracket.
+        let columnTypePattern = @"^[^[(\n]*" //@"[^\[(]*(?= [\[\(])"
+        let namePattern = @"(?<= \[).*(?=[\]])"
+        // [\w-]+ = Pattern matches any letter, number, underscore and minus, but minimum 1
+        // (:|_) = ':' or '_', latter is important for url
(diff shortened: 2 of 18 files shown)


Freymaurer:
This will brick with swate i think, so until i have updated my regex we should not expect this for production. Plus we really need to unify all necessary regex pattern. Can you move them into an extra (and externally accessible) module? With unit tests? (I can do the unit tests if you want)

HLWeil:
@Freymaurer, Currently the library only has Assay file `reading` capabilities anyways, and this does work with Swate either way. 

 I'm not sure about the external module thing. IMO this adds complexity to the project structure that is neither intuitive nor necessary. The problem atm is, that you can't use the `Fable` compiled ISADotNet in Swate, right? 

For this to work, using a different Json Serializer would be the solution. I think time would better be invested in that.

HLWeil:
Perfect, thank you!

I will merge this now, if there aren't any other objections 😄

PR #76, opened 2023-01-26
Improvements to querymodel

HLWeil:
There are now actual `QNode` objects that can be directly asked for their associated nodes and values:

E.g:
```fsharp
for sample in ps.LastSamples() do
    !! sample.Values.["Organism"].TryValueText
```
instead of 
```fsharp
for sample in ps.LastSamples() do
    !! ps.ValuesOf(sample).["Organism"].TryValueText
```

Also now Linq style querying supported:
```fsharp
for sample in ps.LastSamples() do
  cells {
    whereName "Organism"
    selectValueText
    required
  }
```


HLWeil:
Tests fail because json schemas cannot be downloaded for validation for whatever reason. Works fine locally and doesn't affect anything. Will merge anyways...

Issue #90, opened 2023-05-08
[Fable] setup to-do


Freymaurer:
- [ ] Test full release chain
  - [x] successful tests
  - [ ] bundle as npm package
  - [ ] nuget package
  - [ ] Automated release to both npm and nuget? Only as flag?
- [x] github action tests on commit to main or PR to main
- [ ] Code coverage? 
  - [ ] dotnet
  - [ ] js?
- [x] make package.json include both `ISADotNet` and `ISADotNet.Json`
https://github.com/nfdi4plants/ISADotNet/blob/35539e7a7a8d2d5acf1b9a647b933c5120f9c5c7/package.json#L7

Freymaurer:
Does it make sense to work on the open tasks when the possible repo restructuring is in sight? @kMutagene @omaus @HLWeil

HLWeil:
I think working on the packaging and stuff doesn't really make sense. But it would definitely be a possibility to finish up the fable transpilation of the ISADotNet.XLSX project. Then we could be sure that the 3 core libraries are fable-ready.

kMutagene:
@Freymaurer setting up / getting to know reproducible workflows for test coverage in both .NET and js would be something we can always benefit from, actually implementing stuff should wait imo (other than what @HLWeil mentioned)

PR #91, opened 2023-05-08
Fable

Freymaurer:
Finish basic project setup and ISADotNet compatibility together with @HLWeil for [Fable Milestone](https://github.com/nfdi4plants/ISADotNet/milestone/1) 🎉


HLWeil:
Yay!
I think regarding your to-do (#90), getting the CI up and running would be priority number one. So that we can better keep track on the state of the repo. Feel free to merge, when you see that green check mark ✅

PR #92, opened 2023-05-09
Json IO rework

HLWeil:
## Library
- moved json parsing into a new `ISADotNet.Json` project
- restructured everything for use with `Thoth.Json`
- removed file system access of json io functions
## Tests
- reorganized json tests into own repo, together with validation
- moved test files into assembly so test functions don't need file system access


package.json
Freymaurer, line 1: Why remove the tests from main proj?

src/ISADotNet.Validation/JsonSchemaValidation.fs
```diff
@@ -2,15 +2,20 @@
 
 module JSchema = 
    
+#if FABLE_COMPILER
+    let validate (schemaURL : string) (objectString : string) =
```
Freymaurer, line 6: Open issue for this

package.json
HLWeil: Wanted to test `ISADotNet.Json` but wasn't sure how to append these instead of replacing the original ones. Noted this in the to-do

src/ISADotNet.Validation/JsonSchemaValidation.fs
HLWeil: #90 added this here

PR #96, opened 2023-06-01
JSON-LD export and corresponding Tests

floWetzels:
Added json-ld export functions for all ISA types as well as id generators. The new export makes use of extended converter options.


HLWeil (approved):
Looks good 👍

PR #110, opened 2023-06-13
[WIP] Arctrl swate api

Freymaurer:
Again, so you are always up to date @HLWeil


src/ISA/ISA/ARCTypes/ARCTable.fs
```diff
@@ -1,12 +1,69 @@
...
+        let nNewColumn = columns.Length
+        let columnHeaders, columnRows = columns |> Array.map (fun x -> x.Header, x.Cells) |> Array.unzip
+        let nextHeaders = 
+            let ra = ResizeArray(this.ValueHeaders)
+            ra.InsertRange(index, columnHeaders)
+            ra.ToArray()
+        let maxRows = 
+            let maxRows = columnRows |> Array.map (fun x -> x.Length) |> Array.max
+            System.Math.Max(this.RowCount, maxRows)
+        let nextBody =
```
HLWeil, line 39: Should we make the members mutable instead? This could boost performance and would probably follow expectations (See behaviour of Deedle, where member do mutable operations on the frame. Static members work immutable there.).

@muehlhaus

src/ISA/ISA/ARCTypes/ARCTable.fs
```diff
@@ -1,12 +1,69 @@
...
+    member this.addColumns ( columns: CompositeColumn [], ?index: int) = 
+        let index = Option.defaultValue this.ColumnCount index
+        columns |> Array.iter (fun x -> x.validate(true) |> ignore)
+        let nNewColumn = columns.Length
+        let columnHeaders, columnRows = columns |> Array.map (fun x -> x.Header, x.Cells) |> Array.unzip
+        let nextHeaders = 
+            let ra = ResizeArray(this.ValueHeaders)
+            ra.InsertRange(index, columnHeaders)
+            ra.ToArray()
+        let maxRows =
```
HLWeil, line 36: Add to and use static properties of the table instead?

src/ISA/ISA/ARCTypes/ARCTable.fs
```diff
@@ -1,12 +1,69 @@
...
+    static member init(name: string) = {
+        Name = name
+        ValueHeaders = [||]
+        Values = System.Collections.Generic.Dictionary<int*int,CompositeCell>()
+    }
+
+    member this.ColumnCount = this.ValueHeaders.Length
+    member this.RowCount = this.Values.Keys |> Seq.maxBy snd |> snd
+
+    member this.addColumn (header:CompositeHeader, ?cells: CompositeCell [], ?index: int) =
```
HLWeil, line 22: `add(?index)`
or 
```
append()
insert(index)
```
?

@muehlhaus

src/ISA/ISA/ISA.fsproj
HLWeil, line 1: Noticed this too...
`ArcTypes` or `ARCTypes`?
Pascalcase is probably more readable so I agree with your solution

src/ISA/ISA/ARCTypes/ARCTable.fs
```diff
@@ -1,12 +1,69 @@
...
     }
 
+    /// Create ArcTable with empty 'ValueHeader' and 'Values' 
+    static member init(name: string) = {
+        Name = name
+        ValueHeaders = [||]
+        Values = System.Collections.Generic.Dictionary<int*int,CompositeCell>()
+    }
+
+    member this.ColumnCount = this.ValueHeaders.Length
```
HLWeil, line 19: I would replace this with a getter to be sure for future changes. IF we make the table mutable, this construct here would not update with the table.

src/ISA/ISA/ARCTypes/ARCTable.fs
HLWeil: Just noticed the "Add to" was already done above

HLWeil (commented):
We could potentially merge and move my comments into a design issue or settle this beforehand. What do you think, @Freymaurer?

src/ISA/ISA/ARCTypes/ARCTable.fs
Freymaurer: To my surprise it does not allow using both curried and tupled arguments between static members and members:

image

So this is actually a decision we need to make. static members without optional parameters or no f# piping.

This also relates to: https://github.com/nfdi4plants/ISADotNet/pull/110#discussion_r1228289834

src/ISA/ISA/ARCTypes/ARCTable.fs
Freymaurer: Can you give an example?

HLWeil:
```fsharp
    member this.ColumnCount 
        with get() = this.ValueHeaders.Length
```

src/ISA/ISA/ARCTypes/ARCTable.fs
HLWeil: Even for different capitalization?

Freymaurer: 🤕 wow, i am sorry

image

src/ISA/ISA/ARCTypes/ARCTable.fs
Freymaurer: never used this before. i fully trust you witth this 👍

Freymaurer:
Tests are running for dotnet and fable mocha! Think this is ready to merge!

PR #115, opened 2023-06-21
Rework spreadsheet parsing

HLWeil:
- ARCtrl does not have direct IO access so the Spreadsheet parsers now only allows transformations between the datamodel and the in-memory spreadsheet 
- Investigation file parser now uses FsSpreadsheet instead of Documentformat.OpenXML 
- Parsers should now use string parsers defined in the base ISA project
- Parsers should now return the newly defined `ArcTypes` based on `ArcTable` instead of `Process`

To-Do
- [x] Add transformers between `ArcTable` and `Protocol`
- [ ] Add transformers between `ArcTable` and `Process`
- [x] Finish up ArcTable parser
  - [x] Add all headers
  - [ ] React to different value combinations based on header
- [ ] Add comment to every function

- [ ] Keep or remove FSharpAux?


src/ISA/ISA.Spreadsheet/ArcAssay.fs
```diff
@@ -0,0 +1,88 @@
+﻿module ISA.Spreadsheet.ArcAssay
+
+open ISA
+open FsSpreadsheet
+
+let obsoloteAssaysLabel = "ASSAY METADATA"
```
Freymaurer, line 6: Shoould this be written up modular? Like is "METADATA" used somewhere else too? How about "PERFORMERS"? 

If so, then this should just be `assaysLabel` + "metadataLabel"

src/ISA/ISA.Spreadsheet/ArcAssay.fs
```diff
@@ -0,0 +1,88 @@
...
+                assays |> Seq.tryHead 
+                |> Option.defaultValue (ArcAssay.create()) 
+                |> ArcAssay.setPerformers (Option.fromValueWithDefault [] contacts)
+        
+        if en.MoveNext () then
+            let currentLine = en.Current |> SparseRow.tryGetValueAt 0
+            loop currentLine [] [] 1
+            
+        else
+            failwith "emptyInvestigationFile"
```
Freymaurer, line 53: assay right? not investigation?

src/ISA/ISA.Spreadsheet/ArcTable.fs
```diff
@@ -0,0 +1,53 @@
+﻿module ISA.Spreadsheet.ArcTable
+
+open ISA
+open FSharpAux
+open FsSpreadsheet
+
+let annotationTablePrefix = "annotationTable"
```
Freymaurer, line 7: This should be unified between Swate, Swate-standalone and ISA.Spreadsheet. 

In addition remember to make all strings like this to literals

src/ISA/ISA.Spreadsheet/ArcTable.fs
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+/// Returns the protocol described by the headers and a function for parsing the values of the matrix to the processes of this protocol
+let tryFromFsWorksheet (sheet : FsWorksheet) =
+    let annotationTable =
+        sheet.Tables
+        |> Seq.tryFind (fun t -> t.Name.StartsWith annotationTablePrefix)
+    match annotationTable with
+    | Some t -> 
+        let compositeColumns =
```
Freymaurer, line 25: This function would be nice to parse Excel tables to ArcTable. So maybe pull it outside for access.

src/ISA/ISA.Spreadsheet/CompositeCell.fs
```diff
@@ -0,0 +1,24 @@
+﻿module ISA.Spreadsheet.CompositeCell
+
+open ISA
+open FsSpreadsheet
+
+let fromFsCells (cells : list<FsCell>) : CompositeCell =
```
Freymaurer, line 6: This is meant to be used on grouped cells? This should be written as `<summary>`. Also implement descriptive failwith message.

src/ISA/ISA.Spreadsheet/CompositeCell.fs
```diff
@@ -0,0 +1,24 @@
...
+    let cellValues = cells |> List.map (fun c -> c.Value)
+    match cellValues with
+    | [v] -> CompositeCell.createFreeText v
+    | [v1;v2;v3] -> CompositeCell.createTermFromString(v1,v2,v3)
+    | [v1;v2;v3;v4] -> CompositeCell.createUnitizedFromString(v1,v2,v3,v4)
+    | _ -> 
+        failwithf "Dafuq"
+
+let toFsCells isTerm hasUnit (cell : CompositeCell) : list<FsCell> =
+    match cell with
```
Freymaurer, line 16: What is happening here? Why is `isTerm`/`hasUnit` given from outside? As this should be coming only from the CompositeCell?

If this must stay here, pls use the `ToCompositeCellType()` functions first to reduce code duplication:
https://github.com/nfdi4plants/ISADotNet/blob/8d174a3178e115aa16a8f4538e225626d68a497c/src/ISA/ISA/ArcTypes/CompositeCell.fs#L26

So have a `TermToFsCell`, `FreetextToFsCell`, `UnitizedToFsCell` subfunction which you call after using the conversion functions referenced above!

src/ISA/ISA.Spreadsheet/CompositeColumn.fs
```diff
@@ -0,0 +1,44 @@
+﻿module ISA.Spreadsheet.CompositeColumn
+
+open ISA
+open FsSpreadsheet
+
+
+let fromFsColumns (columns : list<FsColumn>) : CompositeColumn =
```
Freymaurer, line 7: This again is used after grouping right?

src/ISA/ISA.Spreadsheet/CompositeColumn.fs
```diff
@@ -0,0 +1,44 @@
...
+            |> CompositeCell.fromFsCells
+        |]                 
+    CompositeColumn.create(header,cells)
+
+
+let toFsColumns (column : CompositeColumn) : FsCell list list =
+    let hasUnit = column.Cells |> Seq.exists (fun c -> c.isUnitized)
+    let isTerm = column.Header.IsTermColumn
+    let header = CompositeHeader.toFsCells hasUnit column.Header
+    let cells = column.Cells |> Array.map (CompositeCell.toFsCells isTerm hasUnit)
```
Freymaurer, line 27: Okay, i don't get it. The CompositeCells must know what they are? Why try to tell them from outside?

src/ISA/ISA.Spreadsheet/CompositeColumn.fs
```diff
@@ -0,0 +1,44 @@
...
+    if hasUnit then
```
Freymaurer, line 28: better: 

```fsharp
[
  [header.[0]] @ cells.[0]
  if isTerm || hasUnit then
    [header.[1]] @ cells.[1]
    [header.[2]] @ cells.[2]
  if hasUnit then
    [header.[3]] @ cells.[3]
]
```

src/ISA/ISA.Spreadsheet/CompositeHeader.fs
```diff
@@ -0,0 +1,101 @@
+﻿module ISA.Spreadsheet.CompositeHeader
+
+open ISA
+open FsSpreadsheet
+
+module ActivePattern = 
+
+    open ISA.Regex.ActivePatterns
+
+    let mergeTerms tsr1 tan1 tsr2 tan2 =
```
Freymaurer, line 10: When is this used?

src/ISA/ISA.Spreadsheet/CompositeHeader.fs
```diff
@@ -0,0 +1,101 @@
...
+        if tsr1 <> tsr2 then failwithf "TermSourceRef %s and %s do not match" tsr1 tsr2
+        if tan1 <> tan2 then failwithf "TermAccessionNumber %s and %s do not match" tan1 tan2
+        {|TermSourceRef = tsr1; TermAccessionNumber = tan1|}
+
```
Freymaurer, line 14: This is very confusing, not sure when this is used? Only on headers after grouping?

Are you sure you cannot recycle this function more?

https://github.com/nfdi4plants/ISADotNet/blob/8d174a3178e115aa16a8f4538e225626d68a497c/src/ISA/ISA/ArcTypes/CompositeHeader.fs#L122

src/ISA/ISA.Spreadsheet/CompositeHeader.fs
```diff
@@ -0,0 +1,101 @@
...
+    | Factor f -> f
+    | Characteristic c -> c
+    | Input i -> i
+    | Output o -> o
+    | FreeText ft -> ft
+    | _ -> raise (System.NotImplementedException("parseCompositeHeader"))
+
+let toFsCells (hasUnit : bool) (header : CompositeHeader) : list<FsCell> = 
+    if header.IsSingleColumn then
+        [FsCell(header.ToString())]
```
Freymaurer, line 92: better: 

```fsharp
        [
            FsCell(header.ToString())
            if hasUnit then FsCell("Unit")
            if hasUnit || isTerm then FsCell($"Term Source REF ({header.GetColumnAccessionShort})")
            if hasUnit || isTerm then FsCell($"Term Accession Number ({header.GetColumnAccessionShort})")
        ]
```

src/ISA/ISA.Spreadsheet/CompositeHeader.fs
```diff
@@ -0,0 +1,101 @@
...
+    elif header.IsTermColumn then
+        [
+            FsCell(header.ToString())
+            if hasUnit then FsCell("Unit")
+            FsCell($"Term Source REF ({header.GetColumnAccessionShort})")
```
Freymaurer, line 97: Is this nowhere else used? Should "Term Source REF" not be a string literal, accessible for reuse? Same for "Term Accession Number"

src/ISA/ISA.Spreadsheet/InvestigationFile/Study.fs
```diff
@@ -1,12 +1,11 @@
 namespace ISA.Spreadsheet
 
 open ISA
-open ISA.API
 open Comment
 open Remark
 open System.Collections.Generic
 
-module Study = 
+module Studies =
```
Freymaurer, line 8: Should any of this be usable from js? If yes, please remember to use root modules only.

src/ISA/ISA/ArcTypes/CompositeHeader.fs
```diff
@@ -185,7 +183,7 @@ type CompositeHeader =
         match this with | ProtocolType -> true | anythingElse -> false
 
     /// <summary>
-    /// This function sets the associated term accession for featured columns. 
+    /// This function gets the associated term accession for featured columns.
```
Freymaurer, line 186: Ups, good catch!

src/ISA/ISA/ArcTypes/CompositeHeader.fs
```diff
@@ -194,6 +192,18 @@ type CompositeHeader =
         | ProtocolType -> "DPBO:1000161"
         | anyelse -> failwith $"Tried matching {anyelse} in getFeaturedColumnAccession, but is not a featured column."
 
+    /// <summary>
+    /// This function gets the associated term accession for term columns. 
+    /// </summary>
+    member this.GetColumnAccessionShort =
```
Freymaurer, line 198: change name to `GetAccessionShort`? It is implemented on the header without input. I think adding `Column` makes the name more vague than removing it.

Or rename to `GetHeaderAccessionShort` at least

src/ISA/ISA/Regex.fs
```diff
@@ -65,18 +92,116 @@ module Pattern =
...
+            | "Characteristic" 
+            | "Characteristics"
+            | "Characteristics Value" -> Some r.TermName
+            | _ -> None
+        | _ -> None
+
+    /// Matches a term string (either short or URI) and returns the term source ref and the annotation number strings.
+    /// 
+    /// Example 1: "MS:1003022" --> term source ref: "MS"; annotation number: "1003022"
+    ///
```
Freymaurer, line 153: comment seems outdated! Please Update!

src/ISA/ISA/Regex.fs
```diff
@@ -88,11 +213,11 @@ open System.Text.RegularExpressions
 /// 
 /// **Example 2:** "Term Source REF (NFDI4PSO:0000064)" --> "NFDI4PSO:0000064"
 /// </summary>
-let tryParseTermAccession (str:string) =
+let tryParseTermAnnotation (str:string) =
```
Freymaurer, line 216: Not sure why you renamed this function and all related patterns from `Accession` to `Annotation`, as the information always comes from "Term Accession Number" column in ISA-TAB

tests/ISA/ISA.Tests/ISA.Tests.fsproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="YoloDev.Expecto.TestSdk" Version="[0.13.2]" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="[17.4.0]" />
     <PackageReference Update="FSharp.Core" Version="7.0.0" />
+    <PackageReference Update="FSharp.Core" Version="6.*" />
```
Freymaurer, line 27: Is it good practise to reference two FSharp.Core versions?

Freymaurer (commented):
One final thing, please check how many FSharpAux functions are used. And if deemed possible, only include the necessary ones, to reduce dependencies.

src/ISA/ISA/Regex.fs
HLWeil: The information either comes from `Term Accession Number` or `Term Source REF`. And what you get as a result is accession number `AND` term source ref.

src/ISA/ISA/ArcTypes/CompositeHeader.fs
HLWeil: Agreed, that the `column` gives no useful information here

src/ISA/ISA.Spreadsheet/ArcTable.fs
HLWeil: Any idea about where to put this? I could put this together with the `tryFromFsWorksheet: FsWorksheet -> FsTable option` function you mentioned below into a `AnnotationTable.fs` and module.

HLWeil:
Thanks for the input. I would leave the `FSharpAux` question open to general input first, as I think this is a general question we could find a streamlined solution for.

src/ISA/ISA.Spreadsheet/ArcTable.fs
Freymaurer: Maybe put it on some member of `ArcTable` ? `ArcTable.toAnnotationTableName()` ?

src/ISA/ISA.Spreadsheet/ArcTable.fs
Freymaurer: I actually did mean the stuff creating `compositeColumns`, but good to have the other one too^^

src/ISA/ISA/Regex.fs
```diff
@@ -56,17 +56,17 @@ module Pattern =
 
     /// Hits term accession, without id: ENVO:01001831
     [<LiteralAttribute>]
-    let TermAnnotationShortPattern = @"(?<idspace>\w+?):(?<localid>\w+)" //prev: @"[\w]+?:[\d]+"
```
Freymaurer: This is official vocabulary found in .obo file format specification

src/ISA/ISA.Spreadsheet/ArcTable.fs
HLWeil: Oh :D
It's kinda hard to see in the snippets github creates

src/ISA/ISA/Regex.fs
HLWeil: Can you give a link?

Freymaurer: https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_4.html#S.1.6

src/ISA/ISA.Spreadsheet/ArcTable.fs
HLWeil: Hmm not so sure about this, the step here is basically just a composition of two other functions. But will try to find a sensible name

Freymaurer:
> Add comment to every function

You do not need to do this for every function right now. But there are functions only useful under certain conditions. These should be documented

Issue #138, opened 2023-07-20
[BUG] Metadata sheet ISA terms are sometimes nested, sometimes not


omaus:
**Describe the bug**
In XLSX metadata sheets, terms are sometimes nested and sometimes not:

Nested:
![image](https://github.com/nfdi4plants/ISADotNet/assets/47781170/ee53e001-83de-4ab3-b5b7-c0e9c7968370)
Here: Investigation file: e.g. `INVESTIGATION` -> `INVESTIGATION PUBLICATIONS` -> `Investigation Publication Title`

Not nested:
![image](https://github.com/nfdi4plants/ISADotNet/assets/47781170/f81bfe92-048f-4cf5-922a-baf6011ce3e2)
Here: Assay file: e.g. `ASSAY PERFORMERS` -> `First Name`

**Expected behavior**
Shouldn't this be nested as well in Assay file? Or not nested in Investigation file and Study file?

CC: @kMutagene

kMutagene:
> Shouldn't this be nested as well in Assay file? Or not nested in Investigation file and Study file?

IIRC the investigation file looks exactly the same way that files created by original isa tooling look like. For the study metadata section (which is new in the arc, not included in original isa) we follow this scheme, but break it in the assay file.
 
So i think the assay metadata section should look like this:

```
ASSAY METADATA
Asssay Measurement Type
Asssay Measurement Type Term Accession Number
Asssay Measurement Type Term Source REF
Asssay Technology Type
Asssay Technology Type Term Accession Number
Asssay Technology Type Term Source REF
Asssay Technology Platform
Asssay File Name
ASSAY PERFORMERS
Asssay  Performer Last Name
Asssay  Performer First Name
Asssay  Performer Mid Initials
Asssay  Performer Email
Asssay  Performer Phone
Asssay  Performer Fax
Asssay  Performer Address
Asssay  Performer Affiliation
Asssay  Performer Roles
Asssay  Performer Roles Term Accession Number
Asssay  Performer Roles Term Source REF
```

kMutagene:
For future reference, I found more inconsistencies:

- the metadata sheet names are different: for Investigation, it is `isa_investigation`, while for Study/Assay it is `Study`/`Assay` 
  --> i propose `isa_assay`/`isa_study` for Study/Assay as names for metadata sheets

[...]

PR #140, opened 2023-07-27
#138: add uniform metadata sheet names

kMutagene:
add uniform metadata sheet names for study and assay metadata sheets, adapt parser logic

you do not seem to have excel IO tests, I guess that's because they are hard to translate with JS test logic. So i could not add tests for this.

I also saw that there is no sheet name logic for parsing investigation files, for investigations just the first sheet is taken:

https://github.com/nfdi4plants/ISADotNet/blob/0e05fb4ba8b2d99af68141ffe632f40f36df3a22/src/ISA/ISA.Spreadsheet/InvestigationFile/Investigation.fs#L192-L200

that might be error-prone.


HLWeil:
Tests are planned for this. Currently there are only tests for either metadata or annotation table sheets. But it should easily possible to combine them into a test that parsed a full workbook instead of only the FsWorksheets. Maybe you could include this in this PR?

HLWeil:
Workbook into here:
https://github.com/nfdi4plants/ISADotNet/blob/arctrl/tests/ISA/ISA.Spreadsheet.Tests/TestObjects/Assay.fs
Test into here:
https://github.com/nfdi4plants/ISADotNet/blob/arctrl/tests/ISA/ISA.Spreadsheet.Tests/AssayFileTests.fs

Sample Annotation Table sheet(s) can be taken from here:
https://github.com/nfdi4plants/ISADotNet/blob/arctrl/tests/ISA/ISA.Spreadsheet.Tests/ArcTableTests.fs

HLWeil (approved):

HLWeil:
- Streamlined Study Parsing with Assay parsing. Now always returns an object where the identifier is autogenerated if missing.
- Study Metadata sheet was reduced to what is the status quo since ArcCommander 0.5.0.
- Reincluded tests

PR #141, opened 2023-07-27
Make ARC types json parseable

### LINKED ISSUE #101: [Feature Request] ISA Json on new data types
- [x] Add ArcTable.toProtocols and ArcTable.fromProcesses functions
- [x] Create Json IO functions for new data types

HLWeil:
Currently, the json reasers/writers only work on the original isa-json types. To make them also work on our ArcTable/ArcAssay/ArcStudy/ArcInvestigation types. We need converters between the two.

For this I added in this PR functions and tests to translate between `isa-tab ArcTables` and `isa-json Processes`.

Still to-do:
- [x] Assay,Study,Investigation converters
- [x] Assay,Study,Investigation converter tests
- [x] ArcAssay,ArcStudy,ArcInvestigation json IO

@Freymaurer 
I had to do some changes to the `Identifier` modules, as I needed this functionality in the translation between ArcAssays and Assays. Could you take a look into this?

closes #101


src/ISA/ISA/ArcTypes/ArcInvestigation.fs
```diff
@@ -266,4 +266,37 @@ type ArcInvestigation(identifier : string, ?title : string, ?description : strin
         )
 
 
-    
\ No newline at end of file
+    /// Transform an ArcInvestigation to an ISA Json Investigation.
+    member this.ToStudy() : Investigation =
```
Freymaurer, line 270: This is wrong?

Freymaurer: Should these be type extensions in a different name space ?

src/ISA/ISA/ArcTypes/CompositeHeader.fs
```diff
@@ -114,6 +114,23 @@ type CompositeHeader =
         | Output io             -> io.asOutput
         | FreeText str          -> str
 
+    member this.ToTerm() =
+        match this with
```
Freymaurer, line 118: Do we need term accessions for these headers?

src/ISA/ISA/ArcTypes/CompositeHeader.fs
```diff
@@ -159,6 +176,16 @@ type CompositeHeader =
         | FreeText s when s.ToLower() = "derived data file" -> true
         | _ -> false   
 
+    /// <summary>
+    /// Is true if this Building Block type is a CvParamColumn.
+    ///
+    /// The name "CvParamColumn" refers to all columns with the syntax "Parameter/Factor/etc [TERM-NAME]".
```
Freymaurer, line 182: What about "Protocol Type" which is a short form for "Parameter [Protocol Type]"

Freymaurer: It also follows the maincolumn-tsr-tan schema

tests/ISA/ISA.Tests/ArcJsonConversionTests.fs
```diff
@@ -0,0 +1,489 @@
...
+
+    ]
+
+
+let main = 
+    testList "ArcJsonConversion" [
+        tests_protocolTransformation
+        tests_arcTableProcess
+        tests_arcTablesProcessSeq
+    ]
```
Freymaurer, line 489: 🚀

tests/ISA/ISA.Tests/Utils.fs
```diff
@@ -25,4 +33,13 @@ let mySequenceEqual actual expected message =
       message i e
   | i,Some a,None ->
     failwithf "%s. Sequence actual longer than expected, at pos %i found item %A."
-      message i a
\ No newline at end of file
+      message i a
+
+module Expect = 
+
```
Freymaurer, line 39: ArcTable has custom equality and should work?

Freymaurer (changes requested):

src/ISA/ISA/ArcTypes/ArcInvestigation.fs
HLWeil: I thought about it. But both the `ArcAssay` and the `Assay` are part of the same namespace, so I guess the transformation between them should also be there.

src/ISA/ISA/ArcTypes/CompositeHeader.fs
HLWeil: You mean for everything, which is not a CvParam?

src/ISA/ISA/ArcTypes/ArcInvestigation.fs
Freymaurer: Bu what do we except the developer user to see? The only reason why we keep the json types in the main namespace are the smaller subtypes such as OntologyAnnotation. For now keep it there.

src/ISA/ISA/ArcTypes/CompositeHeader.fs
HLWeil: Hmm, well it does return true for the `IsTermColumn`. I needed a functionality to specifically check for all the columns that have freely choosable categories.
image

Maybe I could add to the comment, that this does not return true for ProtocolType?

tests/ISA/ISA.Tests/Utils.fs
HLWeil: Yup I saw that too now.. 😅 
Will check out if this works now out of the box.

HLWeil: Ah but there is no info about what is different. The function here tells me where the difference in the table is.

HLWeil:
Finished to-dos.
I would merge now, if it's fine with you @Freymaurer

Issue #161, opened 2023-08-07
[Feature Request] Cleanup OntologyAnnotation


Freymaurer:
Leftover fields such as: `ID` and duplicate information fields such as localid space

Freymaurer:
#163

PR #170, opened 2023-08-11
Rework ontologyannotation localID field

HLWeil:
#161 
- `LocalID` field removed
- Instead there is an `IDInfo` member with both the `IDSpace` and the `LocalID`

Maybe rename to `TANInfo`? @Freymaurer


Freymaurer:
> Maybe rename to TANInfo? @Freymaurer

sounds good to me

Freymaurer (approved):

PR #174, opened 2023-08-14
Make TechnologyPlatform an ontology term

### LINKED ISSUE #162: [Feature Request] Make "Assay Technology Platform" a ontology term
Make it searchable via api and write all information instead of only a string

HLWeil:
And add helper functions to parse to and from string representation in isa-json
closes #162


Freymaurer (changes requested):
All *non basic* tests use only term name in ontology annotation. Might be good to do one with a full OA.

Freymaurer (approved):

PR #177, opened 2023-08-17
Hotfix UpdateFileSystem

HLWeil:
`Arc.UpdateFileSystem()` did not correctly update the filesystem part of the arc according to its ISA part. This was because some erroneous handling of mutable values.
This PR hotfixes this behaviour and adds some basic tests.

### DIFF
--- a/src/ARCtrl/ARCtrl.fs
+++ b/src/ARCtrl/ARCtrl.fs
@@ -56,20 +56,24 @@ module ARCAux =
 [<AttachMembers>]
 type ARC(?isa : ISA.ArcInvestigation, ?cwl : CWL.CWL, ?fs : FileSystem.FileSystem) =
 
-    let mutable cwl = cwl
-    let mutable fs = 
+    let mutable _isa = isa
+    let mutable _cwl = cwl
+    let mutable _fs = 
         fs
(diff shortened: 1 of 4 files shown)


src/ARCtrl/ARCtrl.fs
```diff
@@ -56,20 +56,24 @@ module ARCAux =
...
-    let mutable fs = 
+    let mutable _isa = isa
+    let mutable _cwl = cwl
+    let mutable _fs = 
         fs
         |> Option.defaultValue (FileSystem.create(FileSystemTree.Folder ("",[||])))
         |> ARCAux.updateFSByISA isa
         |> ARCAux.updateFSByCWL cwl
 
-    member val ISA : ArcInvestigation option = isa with get, set
```
Freymaurer, line 66: Is this change communicated Jonas? It was done to try fix an mutable issue in javascript. Please get some information AND unit test in native js from him.

HLWeil: @JonasLukasczyk,
could you add a `native js` test [here](https://github.com/nfdi4plants/ARCtrl/tree/main/tests/JavaScript) which checks whether the access to the `ISA` field of the `ARC` object works as expected and it can be `mutably set`?

PR #194, opened 2023-09-20
Restructure testing environment

HLWeil:
- Created project `tests/TestingUtils`
- Moved all `testing helper functions` and `test objects` into this project
- Added a reference to this project in all test projects


tests/FileSystem/paket.references
```diff
@@ -1,6 +1,3 @@
 FSharp.Core
-Fable.Core
-Expecto
-Fable.Mocha
 Microsoft.NET.Test.Sdk
```
Freymaurer, line 2: Why not remove these two dependencies?

Freymaurer (approved):

tests/FileSystem/paket.references
HLWeil: Otherwise they don't appear in the test explorer in vs

PR #200, opened 2023-09-20
Solve Issues

### LINKED ISSUE #190: [Feature Request] Add convenience function to access the text value of a CompositeCell
**Is your feature request related to a problem? Please describe.**
See https://github.com/nfdi4plants/ARCtrl/pull/189/files#r1323416055 , just tracking here so it is not forgotten.

### LINKED ISSUE #197: [Feature Request] Creating CompositeHeader on JS side needs API support

### LINKED ISSUE #198: [BUG] ArcTable.UpdateHeader forceConvert flag updates cells not smart enough
```
member this.UpdateHeader (index:int, newHeader: CompositeHeader, ?forceConvertCells: bool) =
        let forceConvertCells = Option.defaultValue false forceConvertCells
        ArcTableAux.SanityChecks.validateColumnIndex index this.ColumnCount false
        /// remove to be replaced header, this is only used to check if any OTHER header is of the same unique type as column.Header
        /// MUST USE "Seq.removeAt" to not remove in mutable object!
        let otherHeaders = this.Headers |> Seq.removeAt index
        ArcTableAux.SanityChecks.validateNoDuplicateUnique newHeader otherHeaders
        let c = { this.GetColumn(index) with Header = newHeader }
        // Test if column is still valid with new header, if so insert header at index
        if c.validate() then
            let setHeader = this.Headers.[index] <- newHeader
            ()
        // if we force convert cells, we want to convert the existing cells to a valid cell type for the new header
        elif forceConvertCells then
            let convertedCells =
                match newHeader with
                | isTerm when newHeader.IsTermColumn -> c.Cells |> Array.map (fun c -> c.ToTermCell())
                | _ -> c.Cells |> Array.map (fun c -> c.ToFreeTextCell())
            this.UpdateColumn(index, newHeader, convertedCells)
        else
            failwith "Tried setting header for column with invalid type of cells. Set `forceConvertCells` flag to automatically convert cells into valid CompositeCell type."
```

Freymaurer:
- closes #198 
- closes #190 
- closes #197

### DIFF
--- a/src/ISA/ISA/ArcTypes/CompositeHeader.fs
+++ b/src/ISA/ISA/ArcTypes/CompositeHeader.fs
@@ -105,6 +105,30 @@ type CompositeHeader =
         Microsoft.FSharp.Reflection.FSharpType.GetUnionCases(typeof<CompositeHeader>) 
         |> Array.map (fun x -> x.Tag, x.Name)
 
+    /// <summary>
+    /// This function is used to programmatically create `CompositeHeaders` in JavaScript. Returns integer code representative of input type.
+    ///
+    /// 0: Expects no input
+    ///
+    /// 1: Expects OntologyAnnotation as input
+    ///
+    /// 2: Expects IOType as input
+    ///
+    /// 3: Expects string as input
+    /// </summary>
+    /// <param name="inp">Can be accessed from `CompositeHeader.Cases`</param>
+    static member jsGetColumnMetaType(inp:int) =
+        match inp with
+        // no input
+        | 4 | 5 | 6 | 7 | 8 | 9 | 10 -> 0
+        // OntologyAnnotation as input
+        | 0 | 1 | 2 | 3 -> 1
+        // iotype as input
+        | 11 | 12 -> 2
+        // string as input
+        | 13 -> 3
+        | anyElse -> failwithf "Cannot assign input `Tag` (%i) to `CompositeHeader`" anyElse
+
     override this.ToString() =
         match this with
         | Parameter oa          -> $"Parameter [{oa.NameText}]"
--- a/src/ISA/ISA/ArcTypes/CompositeHeader.fs
+++ b/src/ISA/ISA/ArcTypes/CompositeHeader.fs
@@ -25,6 +25,10 @@ type IOType =
         Material
     |]
 
+    static member Cases = 
+        Microsoft.FSharp.Reflection.FSharpType.GetUnionCases(typeof<IOType>) 
+        |> Array.map (fun x -> x.Tag, x.Name)
+
     member this.asInput = 
         let stringCreate x = $"Input [{x.ToString()}]"
         match this with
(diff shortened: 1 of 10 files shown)


tests/TestingUtils/Library.fs
```diff
@@ -45,41 +45,39 @@ module Result =
...
         | Error m -> m
 
-
-
 /// Fable compatible Expecto/Mocha unification
 module Expect = 
     open Utils
 
     let inline equal actual expected message = Expect.equal actual expected message
-    let inline notEqual actual expected message = Expect.notEqual actual expected message
```
HLWeil, line 55: Why? This caused errors for me previously.

HLWeil: In js

src/ISA/ISA/ArcTypes/CompositeHeader.fs
```diff
@@ -97,6 +101,34 @@ type CompositeHeader =
...
+    /// 0: Expects no input
+    ///
+    /// 1: Expects OntologyAnnotation as input
+    ///
+    /// 2: Expects IOType as input
+    ///
+    /// 3: Expects string as input
+    /// </summary>
+    /// <param name="inp">Can be accessed from `CompositeHeader.Cases`</param>
+    static member jsGetColumnMetaType(inp:int) =
```
HLWeil, line 120: Sry, I don't quite get this. Could you further describe this in the issue?

HLWeil (approved):

tests/TestingUtils/Library.fs
Freymaurer: inline decreases performance, and it is only required for equal.

src/ISA/ISA/ArcTypes/CompositeHeader.fs
Freymaurer: programmatically decide how to create a specific CompositeHeader while looping over "Cases"

```js
it('jsGetColumnMetaType', function () {
        let cases = CompositeHeader.Cases
        let oa = OntologyAnnotation.fromString("My OA Name")
        let iotype = new IOType(0, [])
        let stringExample = "My Example"
        for (let mycase of cases) {
            let tag = mycase[0]
            let code = CompositeHeader.jsGetColumnMetaType(tag)
            switch (code) {
                case 0:
                    let header1 = new CompositeHeader(tag, [])
                    equal((header1.IsSingleColumn || header1.IsFeaturedColumn), true);
                    break;
                case 1:
                    let header2 = new CompositeHeader(tag, [oa])
                    equal(header2.IsTermColumn, true);
                    break;
                case 2:
                    let header3 = new CompositeHeader(tag, [iotype])
                    equal(header3.IsIOType, true);
                    break;
                case 3:
                    let header4 = new CompositeHeader(tag, [stringExample])
                    equal(header4.isFreeText, true);
                    break;
            }
        }
    })
```

Issue #250, opened 2023-11-10
[BUG] Deleting assay from ARC will produce faulty write contracts


Freymaurer:
Steps to reproduce:

1. Create ARC with investigation. 
2. Add Assay
3. Remove Assay
4. UpdateFS()
5. GetWriteContracts
6. Returns Write contract "Plaintext" for removed assay.

UpdateFS does not handel "remove" only "add". In the current implementation of "registerable" assays and studies, this behaviour might be intended, but the write contracts are faulty. We would never want to give write contracts for xlsx files.

In addition a lot of file contracts should be changed from WRITE to UPDATE, as we would not want to replace existing .gitignores, instead we want to initialize them with WRITE. In contrast we want to UPDATE registered isa files.

Freymaurer:
@JonasLukasczyk

Freymaurer:
Current solution is removing from ARC.ISA, while also deleting from fs and then rereading the ARC. This can of course never be a permanent solution

Freymaurer:
We now have RemoveAssay/RemoveStudy on ARC level with contracts as return value

PR #257, opened 2023-11-19
Various read write fixes

HLWeil:
- Reading ISA from Contracts will now not return Assay and Study objects, when they are only registered in the investigation file, but have no corresponding folder/file in the filesystem.
- added a ARC.setFileSystem method, #250 
- investigation spreadsheet writer previously failed when it contained a study that was only registered, but no corresponding study object
- reworked test objects for full arc read-in


src/ISA/ISA.Spreadsheet/ArcInvestigation.fs
```diff
@@ -194,7 +194,8 @@ module ArcInvestigation =
             yield  SparseRow.fromValues [contactsLabel]
             yield! Contacts.toRows (Some contactsLabelPrefix) (List.ofArray investigation.Contacts)
 
-            for study in investigation.RegisteredStudies do
+            for studyIdentifier in investigation.RegisteredStudyIdentifiers do
```
Freymaurer, line 197: What exactly is happening here? Can you add a few comments?

Freymaurer (commented):
I want to add ARC level api for removing assay/study, which will return a real REMOVE contract and handle fs updating internal. Let me know what you think about it.

src/ARCtrl/ARC.fs
```diff
@@ -71,6 +71,25 @@ type ARC(?isa : ISA.ArcInvestigation, ?cwl : CWL.CWL, ?fs : FileSystem.FileSyste
...
+    member this.RemoveAssay(assayIdentifier: string) =
+        if this.ISA.IsNone then failwith "Cannot remove assay from null ISA value."
+        this.ISA.Value.RemoveAssay(assayIdentifier)
+        let paths = this.FileSystem.Tree.ToFilePaths()
+        let assayFolderPath = Path.getAssayFolderPath(assayIdentifier)
+        let filteredPaths = paths |> Array.filter (fun p -> p.StartsWith(assayFolderPath) |> not)
+        this.SetFilePaths(filteredPaths)
+        Contract.createDelete(assayFolderPath)
+
+    member this.RemoveStudy(studyIdentifier: string) =
```
HLWeil, line 85: Also unregister?
Maybe we could do the naming like in the ArcCommander:

Delete -> Remove Object from filesystem by removing folder and file
Unregister -> Remove Registry reference from investigation file
Remove -> Delete + Unregister

HLWeil: Nvm, very good

src/ARCtrl/ARC.fs
```diff
@@ -71,6 +71,25 @@ type ARC(?isa : ISA.ArcInvestigation, ?cwl : CWL.CWL, ?fs : FileSystem.FileSyste
 
     member this.FileSystem 
         with get() = _fs
+        and set(fs) = _fs <- fs
+
+    member this.RemoveAssay(assayIdentifier: string) =
```
HLWeil, line 76: These functions should contain update contracts for all registering objects (study,investigation)

Additional helpers:
- ContractCreator Type extensions for ArcInvestigation, ArcStudy and ArcAssay in Contract namespace

PR #263, opened 2023-11-23
Registration sensitive arc type getter functions

HLWeil:
New Keywords: 
 - `Vacant` = RegisteredButNotExisting
 - `Unregistered` = ExistingButNotRegistered

Changes:
- `RegisteredXCount` now only returns the count of those `X` that are registered AND existing, like `RegisteredXs`. 
- `RegisteredXIdentifierCount` returns that what `RegisteredXCount` used to return

Tests

### DIFF
--- a/src/ISA/ISA/Helper.fs
+++ b/src/ISA/ISA/Helper.fs
@@ -30,13 +30,6 @@ module ResizeArray =
             if f i then b.Add(i)
         b
 
-    let exists f (a : ResizeArray<_>) =
-        let mutable state = false
-        for i in a do
-            if f i then
-                state <- true
-        state
-
     let fold f s (a : ResizeArray<_>) =
         let mutable state = s
         for i in a do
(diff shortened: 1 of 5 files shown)


src/ISA/ISA/Helper.fs
```diff
@@ -30,6 +30,13 @@ module ResizeArray =
             if f i then b.Add(i)
         b
 
+    let exists f (a : ResizeArray<_>) =
```
Freymaurer, line 33: This returns true or false and should be done using Seq.exists

Freymaurer (changes requested):
Remove ResizeArray.exists then it is ready to merge!

src/ISA/ISA/Helper.fs
HLWeil: Ah oops

Freymaurer (approved):

PR #284, opened 2023-12-20
Add full conversions for ArcTypes to and from JSON

Freymaurer:
- Add modules with encoders, decoders.
- Add AutoOpen type extensions.
- NOT breaking previous api syntax.
- moving ArcAssay, ArcStudy and ArcInvestigation functions to new files.
- Adding tests ✅

### DIFF
--- a/src/ISA/ISA.Json/Assay.fs
+++ b/src/ISA/ISA.Json/Assay.fs
@@ -89,21 +89,3 @@ module Assay =
 
     //let toFile (path : string) (p:Assay) = 
     //    File.WriteAllText(path,toString p)
-
-
-module ArcAssay = 
-
-    open Assay
-
-    let fromJsonString (s:string) = 
-        GDecode.fromJsonString (decoder (ConverterOptions())) s
-        |> ArcAssay.fromAssay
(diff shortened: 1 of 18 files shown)


src/ISA/ISA.Json/Assay.fs
```diff
@@ -89,21 +89,3 @@ module Assay =
 
     //let toFile (path : string) (p:Assay) = 
     //    File.WriteAllText(path,toString p)
-
-
-module ArcAssay =
```
HLWeil, line 94: Now the default case for exporting an assay is the non isa-json conforming variant. Not sure about this.

HLWeil: E.g. the guys from MIAPPE Wizard are already using ArcInvestigation.fromJsonString with isa-json conforming json as input. This would fail now.

HLWeil: I would argue for keeping this function exactly as is and adding the new variants with some disambiguating suffix.

HLWeil (approved):
👍

PR #292, opened 2024-01-08
Speed boost for ArcTable creation and xlsx parsing

HLWeil:
Create, write and read study is now 12x faster. Ontology Annotations now have flat string as `Annotation Value`


Freymaurer (approved):
lgtm, but should be reworked to full sparse matrix as sson as tool stack is in stable release.

Freymaurer (changes requested):
Only one change must be done: You forgot to make the performance test in ISA.Spreadsheet.Tests more strict, by decreasing the allowed runtime.

HLWeil:
> lgtm, but should be reworked to full sparse matrix as sson as tool stack is in stable release.

Yeah, agreed. But when making the changes I noticed that there ARE indeed some decisions to be made so I postponed the rework.

HLWeil:
I increased the strictness. But I think we might need some other method for this, as time elapsed is heavily dependent on the hardware. A quick search did not result in any better solution though.

Freymaurer (approved):

PR #294, opened 2024-01-12
Fix annotation value read :bug:

Freymaurer:
Annotation values are now always written to string instead of any of with number option. This change was not correctly adapted in the read for annotationValue.


Freymaurer:
I will leave the PR up but release from this bug fix branch to work on ARCitect, Swate development.

RELEASE_NOTES.md
```diff
@@ -1,3 +1,22 @@
+### 1.0.4+167fae8 (Released 2024-1-15)
```
HLWeil, line 1: Commits for 1.0.2 and 1.0.3 got lost (by me)? 

Did you increase patch version three times here?

HLWeil (approved):
lgtm, fixed the problems in ARCitect?

RELEASE_NOTES.md
Freymaurer: 1.0.2, 1.0.3 and 1.0.4 are all part of this PR and are correctly written to RELEASE_NOTES.md

PR #297, opened 2024-01-17
Extend ArcTable.Join logic 

Freymaurer:
Give option to give index where to join table.

✅  Added tests (3)

### DIFF
--- a/src/ISA/ISA/ArcTypes/ArcTable.fs
+++ b/src/ISA/ISA/ArcTypes/ArcTable.fs
@@ -503,12 +503,17 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
     /// <summary>
     /// This function can be used to join two arc tables.
     /// </summary>
+    /// <param name="index">If not set default to append. -1 will also append.</param>
     /// <param name="table">The table to join to this table.</param>
     /// <param name="joinOptions">Can add only headers, header with unitized cell information, headers with values.</param>
     /// <param name="forceReplace">if set to true will replace unique columns.</param>
-    member this.Join(table:ArcTable, ?joinOptions: TableJoinOptions, ?forceReplace: bool, ?SkipFillMissing) : unit =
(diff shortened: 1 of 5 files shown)


src/ISA/ISA/ArcTypes/ArcTable.fs
```diff
@@ -503,12 +503,17 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
...
     /// </summary>
+    /// <param name="index">If not set default to append. -1 will also append.</param>
     /// <param name="table">The table to join to this table.</param>
     /// <param name="joinOptions">Can add only headers, header with unitized cell information, headers with values.</param>
     /// <param name="forceReplace">if set to true will replace unique columns.</param>
-    member this.Join(table:ArcTable, ?joinOptions: TableJoinOptions, ?forceReplace: bool, ?SkipFillMissing) : unit =
+    member this.Join(table:ArcTable, ?index: int, ?joinOptions: TableJoinOptions, ?forceReplace: bool, ?skipFillMissing) : unit =
         let joinOptions = defaultArg joinOptions TableJoinOptions.Headers
         let forceReplace = defaultArg forceReplace false
+        let skipFillMissing = defaultArg skipFillMissing true
```
HLWeil, line 513: Shouldn't the default value be `false`?

Freymaurer: Don't know, you had it as true:
image

HLWeil: The `not` makes the difference here. You could read my code as follows: `Only if skipFillMissing is set to true, do not fill missing cells`

HLWeil: The default case being `fill missing cells`

Freymaurer: The new one also has `true` with `not`?

Freymaurer: ah wait i think i get it. Because yours defaulted to None, it would not match

HLWeil (approved):

PR #302, opened 2024-01-26
Several performance increases

HLWeil:
## Improve speed of investigation file writer
Writing investigation file with many rows (e.g. 5000 studies) was very slow with quadratic time. Fixed by using the improved `rowWithRange` function form FsSpreadsheet, which skips checking all rows on every call.

## Improve speed of Reading
- Improved fillMissing by roughly a factor of 8
  
image

- Improved addRows, which had a quadratic time dependency because of using `getRowCount` in every step
image

### DIFF
--- a/tests/ISA/ISA.Spreadsheet.Tests/Performance.Tests.fs
+++ b/tests/ISA/ISA.Spreadsheet.Tests/Performance.Tests.fs
@@ -25,15 +25,9 @@ let private tests_Investigation = testList "Investigation" [
         for i = 0 to 1500 do 
             let s = ArcStudy.init($"Study{i}")
             inv.AddRegisteredStudy(s)
-        let timer = Stopwatch()
-        timer.Start()
-        let wb = ArcInvestigation.toFsWorkbook inv
-        timer.Stop()
-        let runtime = timer.Elapsed.Milliseconds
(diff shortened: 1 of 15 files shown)


tests/ISA/ISA.Spreadsheet.Tests/Performance.Tests.fs
```diff
@@ -19,6 +19,24 @@ let private tests_Study = testList "Study" [
...
+            inv.AddRegisteredStudy(s)
+        let timer = Stopwatch()
+        timer.Start()
+        let wb = ArcInvestigation.toFsWorkbook inv
+        timer.Stop()
+        let runtime = timer.Elapsed.Milliseconds
+        let expected = 1000 // this is too high and should be reduced
+
+        Expect.equal (wb.GetWorksheets().Count) 1 "Worksheet count"
+        Expect.isTrue (runtime <= expected) $"Expected conversion to be finished in under {expected}, but it took {runtime}"
```
Freymaurer, line 36: can you abstract this as `Expect.isFastenThan: func1: unit -> unit * func2: unit -> unit`, which runs both functions and fails if func2 is faster than func 1?
Would be much appreciated and could be added to Pyxpecto

Freymaurer (approved):

PR #320, opened 2024-03-12
Update documentation and add python docs

HLWeil:


Freymaurer (commented):
I love it ❤️ But we really need to clean up json functions ™️

Freymaurer (approved):
lgtm!

PR #321, opened 2024-03-13
Jsonld Export with new bioschemas types

floWetzels:
This PR updates the jsonld/ro-crate export such that the output and contexts now use the new bioschemas types LabProcess and LabProtocol. Also, some structural changes were made to the jsonld output. The json decoders were changed accordingly, so that they can now read typical ISA-JSON or the new RO-Crate format. Each data type therefore has a `fromJsonString` and a `fromJsonldString` function.

The tests were adapted accordingly. However, they are not yet finished. The `InvestigationLD` tests need to be updated, since the current expected json strings are still using the old structure and terms. Furthermore, a new `OutputMatchesInputLDtoLD` test was added in the `ProcessLD` test list that reads the new jsonld string, writes it again and then checks for equality. Such tests should be added for all types (the existing ones read normal ISA-JSON and then write jsonld). @HLWeil pleas add/update these tests before merging.

Another part that needs revision are the readers for Material/Sample/Source. Currently, the type is not determined correctly when reading jsonld. The test `ProcessInputLD - Sample - ReaderSuccessSimple` is deactivated therefore.

### DIFF
--- a/src/ISA/ISA.Json/ArcTypes/ArcAssay.fs
+++ b/src/ISA/ISA.Json/ArcTypes/ArcAssay.fs
@@ -103,11 +103,11 @@ module ArcAssay =
 
     /// exports in json-ld format
     let toJsonldString (a:ArcAssay) = 
-        Assay.encoder (ConverterOptions(SetID=true,IncludeType=true)) None (a.ToAssay())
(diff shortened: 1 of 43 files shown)


HLWeil:
Closing in favor of #330

PR #335, opened 2024-04-03
Improve OntologyAnnotation equality

### LINKED ISSUE #268: [BUG] Improve OntologyAnnotation Equality and Hash
Might give wrong results when ShortAnnotation cannot be retrieved.

Freymaurer:
Improve OntologyAnnotation hash and equality handling. Added tests. 

Closes #268

### DIFF
--- a/ARCtrl.sln
+++ b/ARCtrl.sln
@@ -15,6 +15,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		pyproject.toml = pyproject.toml
 		README.md = README.md
 		RELEASE_NOTES.md = RELEASE_NOTES.md
+		setup.cmd = setup.cmd
 	EndProjectSection
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "ARCtrl.CWL", "src\CWL\ARCtrl.CWL.fsproj", "{1CD34A01-D19A-441F-8F3D-6EF69D9DDA8D}"
--- a/setup.cmd
+++ b/setup.cmd
@@ -1,6 +1,9 @@
 @echo off
 TITLE ARCtrl Setup
 
+ECHO Restore .NET tools
+CALL dotnet tool restore
+
 ECHO Install JavaScript Dependencies
 REM npm is a batch file itself. Must use with "CALL" otherwise will exit afterwards.
 CALL npm i
--- a/setup.cmd
+++ b/setup.cmd
@@ -1,0 +1,15 @@
+@echo off
+TITLE ARCtrl Setup
+
+ECHO Install JavaScript Dependencies
+REM npm is a batch file itself. Must use with "CALL" otherwise will exit afterwards.
+CALL npm i
+
+ECHO Setup Python Virtual Environment
+CALL py -m venv .venv
+
+ECHO Install Python Dependencies
+CALL .\.venv\Scripts\python.exe -m pip install -U pip setuptools
+CALL .\.venv\Scripts\python.exe -m pip install poetry
+CALL .\.venv\Scripts\python.exe -m poetry install --no-root
+ECHO DONE!
--- a/src/Core/OntologyAnnotation.fs
+++ b/src/Core/OntologyAnnotation.fs
@@ -134,11 +134,18 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
     override this.GetHashCode() = 
         [|
             HashCodes.boxHashOption this.Name
-            HashCodes.boxHashOption this.TermSourceREF
-            if this.TANInfo.IsSome then
-                HashCodes.hash this.TANInfo
-            else
-                HashCodes.boxHashOption this.TermAccessionNumber
+            match this.TermSourceREF, this.TANInfo with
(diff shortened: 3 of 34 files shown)


ARCtrl.sln
HLWeil, line 1: Maybe mention as alternative in README?

setup.cmd
```diff
@@ -0,0 +1,15 @@
+@echo off
```
HLWeil, line 1: Does `@echo off` potentially obscure some useful information like "what command is called"?

src/Core/OntologyAnnotation.fs
```diff
@@ -131,31 +131,28 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
...
-                other.TermAccessionShort = this.TermAccessionShort
-                ||
-                other.TermAccessionOntobeeUrl = this.TermAccessionOntobeeUrl
-            elif this.Name.IsSome && other.Name.IsSome then
-                other.NameText = this.NameText
-            elif this.TermAccessionNumber.IsNone && other.TermAccessionNumber.IsNone && this.Name.IsNone && other.Name.IsNone then
-                true
-            else 
-                false
+    override this.GetHashCode() =
```
HLWeil, line 134: How does this hash behave in cases, where the underlying Term is the same but once stored as obo url and once as short tan?

src/Core/OntologyAnnotation.fs
```diff
@@ -131,31 +131,28 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
...
-            elif this.TermAccessionNumber.IsNone && other.TermAccessionNumber.IsNone && this.Name.IsNone && other.Name.IsNone then
-                true
-            else 
-                false
+    override this.GetHashCode() = 
+        [|
+            HashCodes.boxHashOption this.Name
+            match this.TermSourceREF, this.TANInfo with
+            | None, Some taninfo -> // if we get taninfo we assume tsr to be inferrable by taninfo
+                HashCodes.hash {|tsr = taninfo.IDSpace; tan = taninfo.IDSpace + ":" + taninfo.LocalID|}
```
HLWeil, line 139: I would make this explicit, i.e. call hashes on the fields and merge them. Record Type hashes behave a bit weird for fable Python

src/Core/Process/Factor.fs
```diff
@@ -32,7 +32,7 @@ type Factor =
 
     /// Get ISATab string entries from an ISAJson Factor object
     static member toString (factor : Factor) =
```
HLWeil, line 34: Should this also be renamed to "toStringOption"

src/Core/Process/MaterialAttribute.fs
```diff
@@ -29,7 +29,7 @@ type MaterialAttribute =
 
     /// Get ISATab string entries from an ISAJson MaterialAttribute object
     static member toString (ma : MaterialAttribute) =
```
HLWeil, line 31: Should this also be renamed to "toStringOption"?

src/Core/Process/ProtocolParameter.fs
```diff
@@ -31,7 +31,7 @@ type ProtocolParameter =
 
     /// Get ISATab string entries from an ISAJson ProtocolParameter object (name,source,accession)
     static member toString (pp : ProtocolParameter) =
```
HLWeil, line 33: Should this also be renamed to "toStringOption"

tests/Core/OntologyAnnotation.Tests.fs
```diff
@@ -0,0 +1,117 @@
+﻿module OntologyAnnotation.Tests
```
HLWeil, line 1: Did you do changes besides moving?

HLWeil (changes requested):

ARCtrl.sln
Freymaurer: mention what? 😅

setup.cmd
Freymaurer: At least not for me

src/Core/OntologyAnnotation.fs
Freymaurer: taninfo uses information after regex parsing. So everything we support with our regex works fine. I also added some tests for this.

src/Core/OntologyAnnotation.fs
Freymaurer: Yeah really not sure about this. I tried to remove the python specific code and it returned strange numbers. But i must check tomorrow in more detail

src/Core/Process/Factor.fs
Freymaurer: I think i named it "toStringObject", but yes we should rename. Not sure to what.

tests/Core/OntologyAnnotation.Tests.fs
Freymaurer: i added all "equal" tests

ARCtrl.sln
HLWeil: Whoops seems like the comment was crooked :D

I meant the `setup.cmd`

Freymaurer: Yes can do!

HLWeil (approved):

PR #346, opened 2024-04-10
Update EXAMPLE_CreateAssayFile.js

xiaoranzhou:
Fix "undefined" in column name by change the property into a function. After the change the column name is correct (Input [Source Name]).

### DIFF
--- a/docs/scripts_js/EXAMPLE_CreateAssayFile.js
+++ b/docs/scripts_js/EXAMPLE_CreateAssayFile.js
@@ -33,7 +33,7 @@ const myAssay = arctrl.ArcAssay.init("MyAssay");
 const growth = arctrl.ArcTable.init("Growth");
 
 // Add input column with one value to table
-growth.AddColumn(arctrl.CompositeHeader.input(arctrl.IOType.source), [arctrl.CompositeCell.createFreeText("Input1")]);
+growth.AddColumn(arctrl.CompositeHeader.input(arctrl.IOType.source()), [arctrl.CompositeCell.createFreeText("Input1")]);
 
 // Add characteristic column with one value
 const oa_species = arctrl.OntologyAnnotation.fromString("species", "GO", "GO:0123456");


Freymaurer:
Good catch! I also found some more issues with the docs and will update them!

xiaoranzhou:
> Good catch! I also found some more issues with the docs and will update them!

Thanks 😄 . The docs are generally very helpful, I will give heads-up when I found something. 

By the way, I use the ARCtrl to make a [SWATE template generation tool](https://xrzhou.com/ena2swate/) by parsing the ENA's checklist.
I use webpack to bundle the ARCtrl's javascript code and the tool is purely clientside and hosted on a GitHub page.
Now I have two questions:
1. Do you think it make sense to integrate this tool in ARCitect or Swate? I mean, I can also create a small set of template and put them into the Swate template repo, but the selectable requirements/recommended/optional terms does offer some flexibility for a  customized Swate template.
2. Is it possible to add a URL routes so with a URL the Swate can open the template xlsx file and start annotation directly?

Freymaurer:
Ah nice! I just checked it out and i might have gotten a bad example but maybe you forgot adding `term source ref` and `term accession number` to the CompositeHeader elements (see the empty brackets): 
image

There is definetly room for such "extensions" we are currently working on a lot of restructuring and more features. But we will discuss when and how we could add interoperability to external tools. Feel free to give some ideas on a generalized approach for this 🙂

xiaoranzhou:
Thanks a lot for looking at the tool and give the helpful feedback. 
I indeed forgot to add the Mandatory terms of ERC000020 to the ontology mapping files. 
The ENA template generation tool parse the ENA checklists and an [ontology mapping](https://docs.google.com/spreadsheets/d/e/2PACX-1vSelJd6uBG8Ri_vqoxawVbCmy4U5Lvc1vqwOFo7dgBz1_yWOuosL46aG9dZdIB7RvuCOzW5Vnr3ytck/pubhtml) file when the webpage is opened.

PR #347, opened 2024-04-10
Update js docs

Freymaurer:


docs/scripts_js/EXAMPLE_CreateAssayFile.js
```diff
@@ -6,21 +6,6 @@
 
 // Import ARCtrl
 import * as arctrl from "@nfdi4plants/arctrl"
-// Alternative import
```
HLWeil, line 9: Sure we should kick this out? This was actually meant as an alternative, not some leftover code

Freymaurer: this is very basic js syntax knowledge, not sure if we should keep and maintain it

HLWeil (approved):

Issue #354, opened 2024-04-30
Create empty .gitattributes during ARC init


JonasLukasczyk:
When creating a new ARC ARCtrl already provides a WriteContract for `.gitignore`. For LFS we also need a WriteContract for `.gitattributes` with empty content (file just needs to be there).

PR #355, opened 2024-04-30
Add gitattributes contract to gitinitcontracts

HLWeil:
#354 

I've left the dto empty as asked for in the issue, but maybe we could add a default rule like `**/dataset/**`. Let me know what you think, @Freymaurer @JonasLukasczyk


Freymaurer:
> #354 
> 
> I've left the dto empty as asked for in the issue, but maybe we could add a default rule like `**/dataset/**`. Let me know what you think, @Freymaurer @JonasLukasczyk 

Maybe a better question for users?

@Brilator

HLWeil:
I included `**/dataset/**` as default rule.

Freymaurer (approved):

PR #380, opened 2024-06-12
Increase robustness against OntologyAnnotation mutability issues

### LINKED ISSUE #378: [BUG] Incorrect handling of new mutable OntologyAnnotation
I assume quite some functions still use OntologyAnnotation in a way, which expects it to be immutable. 
This change was introduced to unify object handling in ARCtrl not only within f# type but also for native js objects.

The following code replicates the specific case i encountered, in which i create a table, add an empty column, then update the oa in one of the columns:

```fsharp
#r "nuget: ARCtrl, 2.0.0-alpha.6.swate"

open ARCtrl

let assay = ArcAssay.init("My Assay")
let table = assay.InitTable("My Table")

table.AddColumn(CompositeHeader.Input IOType.Source, [|for i in 0 .. 4 do CompositeCell.FreeText $"Source {i}"|])

table.AddColumn(CompositeHeader.Component (OntologyAnnotation.create("instrument model", "MS", "MS:424242")))


match table.Values.[(1,0)] with
| CompositeCell.Term oa -> 
  oa.Name <- Some "New Name"
| _ -> failwith "Expected a term"

table
```

### Expected outcome

First cell in second columns has updated name

### Actual outcome

```
-------------
Input [Source Name]     |       Component [instrument model]
Source 0        |       New Name
Source 1        |       New Name
Source 2        |       New Name
Source 3        |       New Name
Source 4        |       New Name
```

HLWeil:
ArcTable.FillMissingCells and Json CompositeCell compressed decoder now Copy CompositeCell to avoid OntologyAnnotation mutability issues.

closes #378

### DIFF
--- a/src/Core/Table/ArcTableAux.fs
+++ b/src/Core/Table/ArcTableAux.fs
@@ -310,12 +310,12 @@ module Unchecked =
                 let rowKeys = Array.map snd col |> Set.ofArray
                 for rowIndex = 0 to rowCount - 1 do
                     if not <| rowKeys.Contains rowIndex then
-                        addCellAt (columnIndex,rowIndex,defaultCell) values
+                        addCellAt (columnIndex,rowIndex,defaultCell.Copy()) values
             // No values existed in this column. Fill with default cells
             | None ->
                 let defaultCell = getEmptyCellForHeader header None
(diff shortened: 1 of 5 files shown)


Freymaurer (approved):
❤️

Issue #324, opened 2024-03-18
[Feature Request] ISA-JSON export and format alignment


xiaoranzhou:
**Is your feature request related to a problem? Please describe.**
The current version of ARCtrl export only short ISA-JSON  
`{
  "identifier": "Facultative-CAM-in-Talinum",
  "title": "Facultative CAM in Talinum triangulare",
  "description": "Drought tolerance is a key factor for agriculture in the 21st century as it is a major determinant of plant survival in natural ecosystems as well as crop productivity. Plants have evolved a range of mechanisms to cope with drought, including a specialized type of photosynthesis termed Crassulacean acid metabolism (CAM). CAM is associated with stomatal closure during the day as atmospheric CO2 is assimilated primarily during the night, thus reducing transpirational water loss. The tropical herbaceous perennial species Talinum triangulare is capable of transitioning, in a facultative, reversible manner, from C3 photosynthesis to weakly expressed CAM in response to drought stress. The transcriptional regulation of this transition has been studied. Combining mRNA-Seq with targeted metabolite measurements, we found highly elevated levels of CAM-cycle enzyme transcripts and their metabolic products in T. triangulare leaves upon water deprivation. The carbohydrate metabolism is rewired to reduce the use of reserves for growth to support the CAM-cycle and the synthesis of compatible solutes. This large-scale expression dataset of drought-induced CAM demonstrates transcriptional regulation of the C3–CAM transition. We identified candidate transcription factors to mediate this photosynthetic plasticity, which may contribute in the future to the design of more drought-tolerant crops via engineered CAM.",
  "submissionDate": "July 07, 2015",
  "publicReleaseDate": "Nov 05, 2015",
  "ontologySourceReferences": [],
  "publications": [],
  "people": [],
  "studies": [],
  "comments": []
}`

 while using  
`toJsonString(arc.ISA)`

**Describe the solution you'd like**
I would like to use ARCtrl to export ISA-JSON use for MARC import.
The ISA-JSON should be similar to the ISA-JSON export generted by ARCcommander or ISAdocNET (in attachment)
The export can be done by using read ARC ans export ISA script (in attachment).
In the best case also aligned with the ISA-JSON generated by VIB (in attachement.)

**Describe alternatives you've considered**
As the ISA object can be seen by using the `console.log(arc.ISA)`, I think only conversions are needed.

**Additional context**
Additional desired behaviors are:
1. Use "@id" to link the "characteristicCategories"

[...]

PR #392, opened 2024-06-25
Make use of @id referencing in isa json writing

HLWeil:
### Description

ISA-json writer functions now contain `useIDReferencing` flag, which will lead to every object in the json file being written only once. For every other appearance of the same object, an object just containing its `@id` will be written in its stead.

#### fsharp
```fsharp
open ARCtrl
open ARCtrl.Json

ArcStudy.toISAJsonString(useIDReferencing = true) study
```

#### javascript
```javascript
import { JsonController } from "./src/ARCtrl/Json.js";

JsonController.Study.toISAJsonString(study, void 0, void 0, true)
```

#### python
```python
from src.ARCtrl.json import JsonController

JsonController.Study().to_isajson_string(study, None, None, True)
```

@Freymaurer @xiaoranzhou 

#324


Freymaurer (approved):
lgtm, nice work!

PR #395, opened 2024-06-26
Rename fix

HLWeil:
- fix and test rename leading to update contracts when the renamed entity is registered and
- fix  getUpdateContracts for adding new assays or studies
- remove isalight:
  Whether studies are written to investigation file and assays are written to study file now solely depends on registration


HLWeil:
fixes https://github.com/nfdi4plants/ARCitect/issues/127

Freymaurer (approved):

PR #403, opened 2024-07-08
#401: add support for `validation_packages.yml`

kMutagene:
This PR
- [x] Adds a type model for validation packages and the `validation_packages.yml` file [according to spec]()
- [x] Adds YAMLicious-based Encoder/Decoder logic
- [x] Adds necessary contracts
~~- [ ] Adds top-level convenience APIs~~
~~- [ ] tests~~
  ~~- [ ] type model~~
  ~~- [ ] read/write yaml~~
  ~~- [ ] contracts~~
  ~~- [ ] API~~

PR #409, opened 2024-07-09
support `validation_packages_yml` (continued)

kMutagene:
This PR continues #403 :

- [x] Adds top-level convenience APIs
- [x] tests
  - [x] type model
  - [x] read/write yaml
  - [x] contracts
  - [x] API


kMutagene:
@Freymaurer finally done adding all those tests 🥇

src/ARCtrl/ARC.fs
```diff
@@ -570,6 +572,39 @@ type ARC(?isa : ArcInvestigation, ?cwl : CWL.CWL, ?fs : FileSystem.FileSystem) =
             ARCtrl.Json.ARC.ROCrate.encoder (Option.get obj.ISA)
             |> ARCtrl.Json.Encode.toJsonString (ARCtrl.Json.Encode.defaultSpaces spaces)
 
+    /// <summary>
+    /// Returns the write contract for the input ValidationPackagesConfig object.
+    ///
+    /// This will mutate the ARC file system to include the `.arc/validation_packages.yml` file if it is not already included.
```
Freymaurer, line 578: Nice comment! ❤️

src/ARCtrl/ARC.fs
```diff
@@ -570,6 +572,39 @@ type ARC(?isa : ArcInvestigation, ?cwl : CWL.CWL, ?fs : FileSystem.FileSystem) =
...
+        if not (Array.contains ValidationPackagesConfigHelper.ConfigFilePath paths) then
+            Array.append [|ValidationPackagesConfigHelper.ConfigFilePath|] paths
+            |> this.SetFilePaths
+        ValidationPackagesConfig.toCreateContract vpc
+
+    /// <summary>
+    /// Returns the delete contract for `.arc/validation_packages.yml`
+    /// </summary>
+    /// <param name="vpc"></param>
+    member this.GetValidationPackagesConfigDeleteContract(vpc) =
```
Freymaurer, line 592: This does not check if it must be removed from filesystem?

src/ValidationPackages/ValidationPackage.fs
```diff
@@ -22,6 +22,15 @@ type ValidationPackage(name, ?version) =
     member this.Copy() =
         ValidationPackage.make this.Name this.Version
 
+    /// Pretty printer 
+    override this.ToString() =
```
Freymaurer, line 26: Maybe do not use yaml style for pretty printing, or use YAMLicious to generate this. But i would prefer typical f# printing syntax

src/ValidationPackages/ValidationPackagesConfig.fs
```diff
@@ -22,15 +22,36 @@ type ValidationPackagesConfig(validation_packages, ?arc_specification) =
     member this.Copy() =
         ValidationPackagesConfig.make this.ValidationPackages this.ARCSpecification
 
-    override this.Equals(obj) =
-        match obj with
-        | :? ValidationPackagesConfig as other_vpc -> other_vpc.ValidationPackages = this.ValidationPackages && other_vpc.ARCSpecification = this.ARCSpecification
+
+    /// Pretty printer 
+    override this.ToString() =
```
Freymaurer, line 27: Same as ValidationPackage.ToString()

src/ValidationPackages/ValidationPackagesConfig.fs
```diff
@@ -22,15 +22,36 @@ type ValidationPackagesConfig(validation_packages, ?arc_specification) =
...
+            if this.ARCSpecification.IsSome then $"arc_specification: {this.ARCSpecification.Value}"
+            "validation_packages:"
+            this.ValidationPackages
+            |> Seq.map (fun vp -> vp.ToString())
+            |> String.concat System.Environment.NewLine
+        ]
+        |> String.concat System.Environment.NewLine
+
+    member this.StructurallyEquals (other: ValidationPackagesConfig) =
+        let sort = Array.ofSeq >> Array.sortBy (fun (vp: ValidationPackage) -> vp.Name, vp.Version)
```
Freymaurer, line 38: Better to use ResizeArray.Sort, as conversions can be performance issue on transpiled code.

src/Yaml/ValidationPackagesConfig.fs
```diff
@@ -8,8 +8,8 @@ module ValidationPackagesConfig =
 
     let encoder (validationpackage : ValidationPackagesConfig) = 
         [
-            "validation_packages", Encode.resizearray ValidationPackage.encoder validationpackage.ValidationPackages
             Encode.tryInclude "arc_specification" Encode.string  (validationpackage.ARCSpecification)
```
Freymaurer, line 11: be better than us and put for example `"validation_packages"` as string Literal. I had to rerelease ARCtrl two times, because i missed updating one key.

Freymaurer (changes requested):
Awesome tests ✅ Requested some minor changes!

src/ValidationPackages/ValidationPackagesConfig.fs
kMutagene: I re-used the pattern introduced by ARCTable here:

https://github.com/nfdi4plants/ARCtrl/blob/dc3822fd1fec14c6009425e618531adf8828119a/src/Core/Table/ArcTable.fs#L697-L702

 I am pretty sure that the person that implemented this encountered the same problem as me when trying to use the `Sort` method of the resize array, as it will produce runtime errors which may be due to the default comparison not working on our types?

Freymaurer: Could be, yes. Alright!

src/ARCtrl/ARC.fs
kMutagene: Was not sure on the logic of this. So we remove the file system entry although the contract has not been fulfilled?

src/Yaml/ValidationPackagesConfig.fs
kMutagene: done, also re-used these literals as expected values in the related yaml tests

src/ARCtrl/ARC.fs
Freymaurer: Yes, as we return the contract we must assume it will be fullfilled. We also do this for RemoveAssay, etc.

kMutagene: okay, done 👍

src/ValidationPackages/ValidationPackage.fs
kMutagene: Also, i think the to string is valuable here because it represents something that only ever exists as a yaml file in the arc.

Freymaurer: It is just that the ToString function should not be responsible for YAML formatting, this should be a different function.

kMutagene: it is not. it is just for printing. formatting for actual yaml export is done with yamlicious.

Freymaurer: Then why do we have tests for this? 😮

Freymaurer (approved):

Freymaurer (approved):

PR #422, opened 2024-07-24
add addColumnFill member function with tests

### LINKED ISSUE #399: [Feature Request] ARCTable AddColumnFill
**Motivation**

Often you want to add a new column to the ArcTable, where every cell value is the same. Instead of having to create an array of identical values before calling `AddColumn`, it would be nice to have a function which fills in the single given value for the length of the table.

**Describe the solution you'd like**

Add an ArcTable member with the name "AddColumnFill" with two required parameters, a header and cell. Adds a new column with the given header and sets this cell for every row of this new column.

The optional parameters should be according to the ones in "AddColumn".

olscholz:
AddColumnFill adds a new column with the given header and sets this cell for every row of this new column. The parameters are equivalent to addColumn. Two testCases have been implemented and two are still under construction. 

closes #399

### DIFF
--- a/src/Core/Table/ArcTable.fs
+++ b/src/Core/Table/ArcTable.fs
@@ -181,12 +181,10 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
             newTable.AddColumn(header, ?cells = cells, ?index = index, ?forceReplace = forceReplace)
             newTable
 
-    // New AddColumnFill adds a new column which fills in the single given value for the length of the table.
-    member this.AddColumnFill (header: CompositeHeader, cell: CompositeCell, ?index: int ,?forceReplace : bool) =  // Member function to add a column with the same value for each cell
-        let cells = Array.init this.RowCount (fun _ -> cell.Copy())    // Create an array of cells with the same value based on the RowCount mutable object so use init instead of Array.create
-        this.AddColumn(header, cells = cells, ?index = index,  ?forceReplace = forceReplace)        // implementing existing AddColumn method
+    member this.AddColumnFill (header: CompositeHeader, cell: CompositeCell, ?index: int ,?forceReplace : bool) =  
+        let cells = Array.init this.RowCount (fun _ -> cell.Copy())    
+        this.AddColumn(header, cells = cells, ?index = index,  ?forceReplace = forceReplace)     
 
-    // Static function to add a column with the same value for each cell (returns a new ArcTable)
     static member addColumnFill (header: CompositeHeader, cell: CompositeCell, ?index: int ,?forceReplace : bool) : ArcTable -> ArcTable =
         fun (table: ArcTable) ->
             let newTable = table.Copy()
--- a/src/Core/Table/ArcTable.fs
+++ b/src/Core/Table/ArcTable.fs
@@ -181,6 +181,18 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
             newTable.AddColumn(header, ?cells = cells, ?index = index, ?forceReplace = forceReplace)
             newTable
 
+    // New AddColumnFill adds a new column which fills in the single given value for the length of the table.
+    member this.AddColumnFill (header: CompositeHeader, cell: CompositeCell, ?index: int ,?forceReplace : bool) =  // Member function to add a column with the same value for each cell
+        let cells = Array.init this.RowCount (fun _ -> cell.Copy())    // Create an array of cells with the same value based on the RowCount mutable object so use init instead of Array.create
+        this.AddColumn(header, cells = cells, ?index = index,  ?forceReplace = forceReplace)        // implementing existing AddColumn method
+
+    // Static function to add a column with the same value for each cell (returns a new ArcTable)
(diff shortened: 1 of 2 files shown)


src/Core/Table/ArcTable.fs
```diff
@@ -181,6 +181,18 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
             newTable.AddColumn(header, ?cells = cells, ?index = index, ?forceReplace = forceReplace)
             newTable
 
+    // New AddColumnFill adds a new column which fills in the single given value for the length of the table.
```
HLWeil, line 184: - Cut out the first "New AddColumnFill".
- Make this a `///` comment

src/Core/Table/ArcTable.fs
```diff
@@ -181,6 +181,18 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
...
+    member this.AddColumnFill (header: CompositeHeader, cell: CompositeCell, ?index: int ,?forceReplace : bool) =  // Member function to add a column with the same value for each cell
+        let cells = Array.init this.RowCount (fun _ -> cell.Copy())    // Create an array of cells with the same value based on the RowCount mutable object so use init instead of Array.create
+        this.AddColumn(header, cells = cells, ?index = index,  ?forceReplace = forceReplace)        // implementing existing AddColumn method
+
+    // Static function to add a column with the same value for each cell (returns a new ArcTable)
```
HLWeil, line 189: - Use the text from the `AddColumnFill` member.
- Make this a `///` comment

tests/Core/ArcTable.Tests.fs
```diff
@@ -1623,6 +1623,46 @@ let private tests_AddColumns =
         ]
     ]
 
+
+//New Test for AddColumnFill
```
HLWeil, line 1627: Cut out the `New`

tests/Core/ArcTable.Tests.fs
```diff
@@ -1623,6 +1623,46 @@ let private tests_AddColumns =
...
+let private tests_AddColumnFill = 
+    testList "AddColumnFill" [ 
+        testCase "addColumnFill to preexisting table with rows" (fun () ->  // AddColumnFill to Table with Rows // create ftestCase which prioritizes the testCase but only on local!!
+            let table = create_testTable()
+            let expectedValue = CompositeCell.createUnitizedFromString("1")
```
HLWeil, line 1632: Rename `expectedValue` to `cell`. This has twofold reason:
- `value` is more often used for values of primitive type
- even though the cell is also used as the `expected` value for the `equal` call later. It's primary use is for being used in the `AddColumnFill` method

tests/Core/ArcTable.Tests.fs
```diff
@@ -1623,6 +1623,46 @@ let private tests_AddColumns =
...
+            table.AddColumnFill(CompositeHeader.Parameter (OntologyAnnotation("chlamy_r")), expectedValue) //or .createTermfromString
+            Expect.equal table.ColumnCount 6 "ColumnCount"
+            Expect.equal table.RowCount 5 "RowCount"
+            let columnIndex = table.ColumnCount - 1
+            let getactualValue rowIndex  = table.Values.[columnIndex,rowIndex]
```
HLWeil, line 1637: I would remove the `getactualValue` function here, as it's underyling body is quite simple anyways. To my perception it's more cluttering.

tests/Core/ArcTable.Tests.fs
```diff
@@ -1623,6 +1623,46 @@ let private tests_AddColumns =
...
+            for ri in 0 .. table.RowCount - 1 do 
+                let value = getactualValue ri 
+                Expect.equal value expectedValue $"RowIndex: {ri}"
+        )
+        ftestCase "addColumnFill to empty table" (fun () -> // "Empty Table" fail if empty flag at least RowCount 1 if table empty etc
```
HLWeil, line 1642: Don't forget to switch from `ftestCase` to `testCase`

HLWeil (changes requested):
When thinking about function names and comment texts, remember that after the PR, the added code is just another part of the code base.
Therefore, the code should not explain itself to a potential reviewer but to a user (of any kind) of the library. Notions like `new function` and such are not necessary in this regard.

tests/Core/ArcTable.Tests.fs
Freymaurer: > [!IMPORTANT]
> We should use `--fail-on-focused-tests` on our CI pipeline!
> 
> https://github.com/Freymaurer/Fable.Pyxpecto?tab=readme-ov-file#focused

Can you do this on your next commit @HLWeil ?

Freymaurer: We actually have multiple options how we want to tackle this:
1. Pass any `<env_variable>` and use `#if <env_variable>` syntax to add this argument in pyxpecto run tests call
2.  propagate --fail-on-focused to pyxpecto and edit workflow

HLWeil (approved):
Changes look good, will merge after tests have run

Issue #433, opened 2024-09-05
[Feature Request] RO Crate Datamodel - ISA Profile: Add typed getter/setter APIs for known fields


kMutagene:
Follow-up for #384 as we decided to have more fine-grained PRs/issues on this.

The ISA profile is implemented as classes, but we still need an ergonomic API to work with them programmatically (e.g. to query properties). 

As per previous discussion with @HLWeil, here are some necessary implementation details:

- all functionality should be implemented as **both** instance methods and static methods on the class
- all methods should use tupled named arguments
- static methods should return a function that calls the instance method of the input.
- Example:
  ```fsharp
  type A(...) =
      ...
      member this.Foo(arg1, arg2) = ...
      static member foo (arg1, arg2) =
          fun (a: A) -> a.Foo(arg1, arg2)
  ```

kMutagene:
@HLWeil what about static members that only have one argument? 

```fsharp
member this.Foo() = ...
static member foo () =
    fun (a: A) -> a.Foo()
```

or 

```fsharp
member this.Foo() = ...
static member foo (a: A) = a.Foo()
```

?

HLWeil:
https://github.com/nfdi4plants/ARCtrl/blob/8b4368aed9a9705a78f95ec752143b1620fb0408/src/Core/Table/ArcTable.fs#L429

Apparently we kept it consistent. Not sure though whether in this case consistency wins over special case usability.

kMutagene:
So this one it is?

```fsharp
member this.Foo() = ...
static member foo () =
    fun (a: A) -> a.Foo()
```

I hate it tbh and would rather change the existing API, but i think this is something for the backlog as of #419

PR #436, opened 2024-09-06
Ro crate data model API

kMutagene:
~~This PR will bundle the work for closing #433~~

We decided to merge this before full completion to unblock other PRs, will continue on another PR


src/ROCrate/ISAProfile/LabProcess.fs
```diff
@@ -24,7 +24,19 @@ type LabProcess(
...
-        DynObj.setOptionalProperty (nameof parameterValue) parameterValue                       this 
-        DynObj.setOptionalProperty (nameof endTime) endTime                                     this 
-        DynObj.setOptionalProperty (nameof disambiguatingDescription) disambiguatingDescription this 
\ No newline at end of file
+        DynObj.setOptionalProperty (nameof executesLabProtocol) executesLabProtocol this
+        DynObj.setOptionalProperty (nameof parameterValue) parameterValue this
+        DynObj.setOptionalProperty (nameof endTime) endTime this
+        DynObj.setOptionalProperty (nameof disambiguatingDescription) disambiguatingDescription this
+
+    member this.GetName() = DynObj.tryGetPropertyValue (nameof name) this |> Option.get
```
HLWeil, line 32: This will cause very unhelpful error messages downstream

src/ROCrate/ROCrateObject.fs
```diff
@@ -39,4 +39,19 @@ type ROCrateObject(id:string, schemaType: string, ?additionalType) =
 
         member this.AdditionalType
             with get() = _additionalType
-            and set(value) = _additionalType <- value
\ No newline at end of file
+            and set(value) = _additionalType <- value
+
+    member this.SetContext (context: #DynamicObj) =
```
HLWeil, line 44: I would argue for creating an "LDContext" or "Context" type which we can reference here.

HLWeil: For now it can just be an otherwise empty class inheriting DynamicObj

HLWeil: Because we will need special behaviour for these objects sometime in the future and it's probably easier if the alias class is already there and referenced where it's used

HLWeil (changes requested):

src/ROCrate/ROCrateObject.fs
kMutagene: Not the biggest fan of adding stubs/unused code/types but will do

src/ROCrate/DynObjExtensions.fs
```diff
@@ -0,0 +1,15 @@
+namespace ARCtrl.ROCrate
+
+open DynamicObj
+
+module DynObj =
+
+    let inline hasProperty (propertyName: string) (obj: #DynamicObj) = DynObj.tryGetPropertyValue propertyName obj |> Option.isSome
+
+    let inline getMandatoryDynamicPropertyOrThrow<'TPropertyValue> (className:string) (propertyName: string) (obj: #DynamicObj) =
```
HLWeil, line 9: Nice 👍

HLWeil (approved):

kMutagene:
continuation in #478

PR #446, opened 2024-10-02
Split ARCtrl project into language-specific project files

### LINKED ISSUE #445: [Feature Request] Separation of ARCtrl into dotnet, js and python
Changes according to https://github.com/thoth-org/Thoth.Json/issues/206

HLWeil:
Removed all `python` and `javascript` specific dependencies from all projects except for top-level ARCtrl.

There, created an `ARCtrl.Javascript` and `ARCtrl.Python` project file, which contain the corresponding dependencies and IO-implementations. 

closes #445 

@MangelMaxime


src/ARCtrl/ARCtrl.Javascript.fsproj
```diff
@@ -0,0 +1,90 @@
...
+    <Compile Include="Json\ROCrateObject.fs" />
+    <Compile Include="Json\ARC.fs" />
+    <Compile Include="WebRequest\WebRequest.Node.fs" />
+    <Compile Include="WebRequest\WebRequest.fs" />
+    <Compile Include="Template.Web.fs" />
+    <Compile Include="ARC.fs" />
+    <Compile Include="Json.fs" />
+    <Compile Include="Xlsx.fs" />
+    <None Include="README.md" />
+  </ItemGroup>
```
MangelMaxime, line 65: If you want to avoid duplication of the code you can move it to `ARCTrl.Common.props` and include it in your different project.

Some of the properties (like Authors, License) can even be moved to the top level `Directory.Build.props`

src/ARCtrl/ARCtrl.Javascript.fsproj
```diff
@@ -0,0 +1,90 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\Contract\ARCtrl.Contract.fsproj" />
+    <ProjectReference Include="..\CWL\ARCtrl.CWL.fsproj" />
+    <ProjectReference Include="..\FileSystem\ARCtrl.FileSystem.fsproj" />
+    <ProjectReference Include="..\Json\ARCtrl.Json.fsproj" />
+    <ProjectReference Include="..\Spreadsheet\ARCtrl.Spreadsheet.fsproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <Content Include="*.fsproj; **\*.fs; **\*.fsi" PackagePath="fable\" />
```
MangelMaxime, line 79: Using [Fable.Package.SDK](https://github.com/fable-compiler/) allows to remove these lines and will asked you to set the `PackageTags` information so your NuGet package can be listed in https://fable.io/packages/

src/ARCtrl/ARCtrl.Javascript.fsproj
HLWeil: Will try, thanks!

src/ARCtrl/ARCtrl.Javascript.fsproj
HLWeil: This did not work for me out of the box, will open an Issue in the given project.

HLWeil: https://github.com/fslaborg/FsSpreadsheet/blob/e59da935ebced63c3bcd7ed9e9dd3e276715fe54/src/FsSpreadsheet/FsSpreadsheet.fsproj#L6

Here are the lines in question. I thought this would be enough to trigger the inclusion of the src files in a Fable folders when running `dotnet pack`

MangelMaxime: It should be enough yes,

Use the following to include the dependency, you didn't have it included by default because it is only done automatically since 1.1.0. This will make your package not include Fable.Package.SDK in its dependencies.

```xml
<PackageReference Include="Fable.Package.SDK">
  <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
  <PrivateAssets>all</PrivateAssets>
</PackageReference>
```

You should also remove:

```xml
<Content Include="*.fsproj; **\*.fs; **\*.fsi" PackagePath="fable\" />
```

If this is still not working I can have a look at it if you want

HLWeil: Hey, sry forgot to write here. The Inclusion of the Source Files works now!

Just got another small issue: https://github.com/fable-compiler/Fable.Package.SDK/issues/6

It does produce a proper nuget file. So ignoring the error kinda works, but would be nice to find the reason behind this.

MangelMaxime: Its strange in theory if there is an error it should fails.

But perhaps MSBuild create the package before and then the CLI exist with an error code. I will check the issue you reported.

MangelMaxime: The issue is caused by this line:

```xml
<PackageTags>ARC F# FSharp dotnet .Net bioinformatics biology fable-library datascience dataplant nfdi metadata</PackageTags>
```

Judging by the content of the tag, I think you wanted to use `<Description>`

HLWeil: > Its strange in theory if there is an error it should fails.
> 
> But perhaps MSBuild create the package before and then the CLI exist with an error code. I will check the issue you reported.

Yeah sry with "ignoring the error" I meant actually catching *and ignoring* the error :D

HLWeil: > The issue is caused by this line:
> 
> ```
> <PackageTags>ARC F# FSharp dotnet .Net bioinformatics biology fable-library datascience dataplant nfdi metadata</PackageTags>
> ```
> 
> Judging by the content of the tag, I think you wanted to use `<Description>`

image

These are actually tags used by nuget to increase findability of the package.

HLWeil: Kinda hotfixed it by merging the two lists (See issue)

HLWeil:
As far as I can tell everything works. I just pushed version 2.1.0-alpha1 to Nuget, npm and pypi.

HLWeil:
So ready to review @Freymaurer @kMutagene 

Thanks a lot for your input, @MangelMaxime!

build/BasicTasks.fs
```diff
@@ -163,6 +163,7 @@ let build = BuildTask.create "Build" [clean] {
         let msBuildParams =
             {p.MSBuildParams with 
                 DisableInternalBinLog = true
+                NodeReuse = false
```
Freymaurer, line 166: Why was this added?

build/PackageTasks.fs
```diff
@@ -104,7 +104,7 @@ module BundlePy =
         run python "-m poetry build" ProjectInfo.pyPkgDir //Remove "-o ." because not compatible with publish 
 
 
-let packPy = BuildTask.create "PackPy" [clean; build; runTests] {
+let packPy = BuildTask.create "PackPy" [clean; build; (*runTests*)] {
```
Freymaurer, line 107: 💀 commented out tests(!)

package.json
Freymaurer, line 1: if you started cleaning this. maybe remove all scripts currently not used (i think we do not use any)

src/ARCtrl/JsonIO/ROCrateObject.fs
Freymaurer, line 1: i am not really a fun of reflection in f# code. I fully trust you in this, that we do not have to rework it in 2 years

src/Package.Metadata.props
Freymaurer, line 1: 🚀❤️

src/ARCtrl/JsonIO/ROCrateObject.fs
HLWeil: Yeah it's always a bit dangerous, but here I think we should be fine, as the `Option` part can be ignored in js and py.

build/BasicTasks.fs
HLWeil: Nevermind, removed it

build/PackageTasks.fs
HLWeil: Whoops, fixed

Issue #407, opened 2024-07-09
[Feature Request] Add config to toggle between datamap file/sheet for contracts


Freymaurer:
in #406 i added logic to create a datamap sheet in assay/study xlsx files. And added read logic to prioritize the datamap sheet.

In the current implementation write/update contracts will always create both sheet and file. How to update this logic is something I and @HLWeil need to discuss.

HLWeil:
On it!

HLWeil:
Closing while https://github.com/nfdi4plants/ARC-specification/issues/134 is ongoing

PR #450, opened 2024-10-08
Update datamap handling in GetUpdateContracts function

HLWeil:
kinda related to #407 

GetUpdateContracts() and GetWriteContracts() functions now have `datamapAsFile` flag. When not set, now the datamap is written as part of the assay/study file. When set to true, it is written as its own file.


HLWeil:
Closing, waiting for https://github.com/nfdi4plants/ARC-specification/issues/134

Issue #447, opened 2024-10-07
[Feature Request] Create accessible spreadsheet parsing functions based on string collections


Freymaurer:
Currently only fsspreadsheet types are supported as input and output:

https://github.com/nfdi4plants/ARCtrl/blob/9af0c63094f3304b6ded3381ec10e1ea72757cfd/src/Spreadsheet/ArcAssay.fs#L19

We require to extract the function internal logic to have something running on string collections

PR #451, opened 2024-10-14
Feature enable sequence based metadata creation

Etschbeijer:
Closses #447

### DIFF
--- a/src/Spreadsheet/ArcAssay.fs
+++ b/src/Spreadsheet/ArcAssay.fs
@@ -76,6 +76,63 @@ module ArcAssay =
         with 
         | err -> failwithf "Failed while parsing metadatasheet: %s" err.Message
 
+    let toMetadataCollection (assay: ArcAssay) =
+        let toRows (assay: ArcAssay) =
+            seq {          
+                yield  SparseRow.fromValues [assaysLabel]
+                yield! Assays.toRows (Some assaysPrefix) [assay]
+
+                yield SparseRow.fromValues [contactsLabel]
+                yield! Contacts.toRows (Some contactsPrefix) (List.ofSeq assay.Performers)
+            }
+        assay
+        |> toRows
+        |> Seq.map (fun row -> SparseRow.getAllValues row)
+
+    let fromMetadataCollection (collection: seq<seq<string option>>) : ArcAssay =
+        try
+            let fromRows (usePrefixes: bool) (rows: seq<SparseRow>) =
+                let aPrefix, cPrefix = 
+                    if usePrefixes then 
+                        Some assaysPrefix, Some contactsPrefix
+                    else None, None
+                let en = rows.GetEnumerator()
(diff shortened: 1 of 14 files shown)


src/Spreadsheet/ArcAssay.fs
```diff
@@ -76,13 +76,70 @@ module ArcAssay =
...
+                yield SparseRow.fromValues [contactsLabel]
+                yield! Contacts.toRows (Some contactsPrefix) (List.ofSeq assay.Performers)
+            }
+        assay
+        |> toRows
+        |> Seq.map (fun row -> SparseRow.getAllValues row)
+
+    let fromMetadataCollection (collection: seq<seq<string option>>) : ArcAssay =
+        try
+            let fromRows (usePrefixes: bool) (rows: seq<SparseRow>) =
```
HLWeil, line 94: Pull this (fromRows) function out to be used both in `fromMetadataSheet` and your function.

src/Spreadsheet/ArcAssay.fs
```diff
@@ -76,13 +76,70 @@ module ArcAssay =
...
+                if en.MoveNext () then
+                    let currentLine = en.Current |> SparseRow.tryGetValueAt 0
+                    loop currentLine [] [] 1
+            
+                else
+                    failwith "empty assay metadata sheet"
+            let rows =        
+                collection 
+                |> Seq.map SparseRow.fromAllValues
+            let hasPrefix =
```
HLWeil, line 128: This hasPrefix part can be moved into the `fromRows` function and dropped as a param there.

HLWeil (changes requested):
Thanks for the PR!

Please reuse the `fromRows` function to reduce code duplication. 

And please undo the styling changes not connected to the specific functions you change, to keep the PR more clean. 
Also I would prefer to keep the `param : type` styling in general. With the `,`, I'm not sure yet. But I heavily advocate against including this kind of cleanups of code-parts unrelated to the PR.

src/Spreadsheet/ArcAssay.fs
Etschbeijer: fixed

src/Spreadsheet/ArcAssay.fs
Etschbeijer: fixed

Etschbeijer:
> Thanks for the PR!
> 
> Please reuse the `fromRows` function to reduce code duplication.
> 
> And please undo the styling changes not connected to the specific functions you change, to keep the PR more clean. Also I would prefer to keep the `param : type` styling in general. With the `,`, I'm not sure yet. But I heavily advocate against including this kind of cleanups of code-parts unrelated to the PR.

Fixed the parameter styling and I will avoid style changes in the futur.

HLWeil (approved):
Thanks, looks good now!

PR #452, opened 2024-10-15
Nuget package fix and test time adjustment

HLWeil:
- Moved references to source files in `JsonIO` folder from `ARCtrl.Common.props` to each project file, as this was causing problems in downstream transpilation (see [Issue in reference project](https://github.com/MangelMaxime/fable-example-multi-targets/issues/1))
- Created an "All" Tests project which references all other tests. This should minimize duplicate transpilation and inrease build time tremendously


Freymaurer:
> This should minimize duplicate transpilation and inrease build time tremendously

~~increase~~ **decrease**

Freymaurer (approved):

Issue #453, opened 2024-10-15
[Feature Request] ARCtrl.Spreadsheet Template.isMetadataSheetName missing


Freymaurer:
The same function exists for assay, study and investigation. Would be nice to keep API unified 🙂

PR #456, opened 2024-10-18
Enable check for metadata sheetname and is meta data sheet for arc template

Etschbeijer:
Closes Issue #453

### DIFF
--- a/src/Spreadsheet/Template.fs
+++ b/src/Spreadsheet/Template.fs
@@ -239,6 +239,12 @@ module Template =
             (ResizeArray tags)  
             lastUpdated
 
+    let isMetadataSheetName (name : string) =
+        name = metaDataSheetName || name = obsoletemetaDataSheetName
+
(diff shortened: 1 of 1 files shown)


Etschbeijer:
> Hey, the Commit (and by this also PR message) could be a bit more concise. Especially `Template` is missing. But otherwise looks fine.
> 
> Also you should extend the logic to match that of ArcAssay and ArcStudy. Making use of the function you created. This way it would also transitively be tested.
> 
> EDIT: Just as @Freymaurer also mentioned

Should be fixed now

HLWeil (changes requested):
Hey, usage of the newly added function in the "fromFsWorkbook" function is still missing. Line 277 including your changes

Etschbeijer:
> Hey, usage of the newly added function in the "fromFsWorkbook" function is still missing. Line 277 including your changes

Do you mean in the tests?

HLWeil:
No, I mean here:

image

HLWeil (approved):

PR #460, opened 2024-10-23
Make OntologyAnnotation work with localID as TAN

### LINKED ISSUE #437: [Feature Request] Parse LocalID OntologyAnnotation Accession Term
https://github.com/nfdi4plants/ARC-specification/pull/120#discussion_r1753830792

```fsharp
#r "nuget: ARCtrl"

open ARCtrl

let oa = OntologyAnnotation("MyOntology", "UO", "0000123")

oa.TermAccessionShort
```

 returns 

```
val it: string = ""
```

but should return

```
val it: string = "UO:0000123"
```

HLWeil:
closes #437

### DIFF
--- a/src/Core/OntologyAnnotation.fs
+++ b/src/Core/OntologyAnnotation.fs
@@ -42,8 +42,14 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
 
     member this.TANInfo = 
         match this.TermAccessionNumber with
-        | Some v -> 
-            Regex.tryParseTermAnnotation v
+        | Some tan -> 
+            match Regex.tryParseTermAnnotation tan with
+            | Some ta -> Some ta
+            | None ->
+                match this.TermSourceREF with
+                | Some "" | None -> None
+                | Some tsr -> 
+                    Some {| IDSpace = tsr; LocalID = tan |}
         | None -> None
 
     member this.NameText = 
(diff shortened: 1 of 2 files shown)


src/Core/OntologyAnnotation.fs
```diff
@@ -42,8 +42,14 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
 
     member this.TANInfo = 
         match this.TermAccessionNumber with
-        | Some v -> 
-            Regex.tryParseTermAnnotation v
+        | Some tan ->
```
Freymaurer, line 45: this should be kind of reverse right?

We have one example the paff(?) subnamespace in the MS ontology. Using this as example we would have:

TSR: `MS`; TAN: `PEFF:0001032`

Then we want to return `{| IDSpace = MS; LocalID = PEFF:0001032 |}`

In my opinion we should give a explicitly set TSR priority.

Freymaurer (changes requested):

src/Core/OntologyAnnotation.fs
HLWeil: But then we would have `MS:PEFF:0001032` as the short-annotation.

Freymaurer: Ahh sorry my bad! Didnt fully understand the function context. All good from my side

Freymaurer (approved):

PR #463, opened 2024-10-23
Template copy and test+fix deepcopying of ArcTable

### LINKED ISSUE #458: [Feature Request] Template.Copy
`Template.Copy`, to unify logic

HLWeil:
closes #458

### DIFF
--- a/src/Core/Table/CompositeHeader.fs
+++ b/src/Core/Table/CompositeHeader.fs
@@ -439,27 +439,7 @@ type CompositeHeader =
         | Characteristic oa -> Characteristic (oa.Copy())
         | Component oa -> Component (oa.Copy())
         | _ -> this
-
-
-
-
-        //| Parameter oa          -> $"Parameter [{oa.NameText}]"
-        //| Factor oa             -> $"Factor [{oa.NameText}]"
-        //| Characteristic oa     -> $"Characteristic [{oa.NameText}]"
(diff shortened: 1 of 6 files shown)


src/Core/Table/CompositeHeader.fs
```diff
@@ -432,6 +432,15 @@ type CompositeHeader =
         | ProtocolType | ProtocolREF | ProtocolDescription | ProtocolUri | ProtocolVersion | Performer | Date | Input _ | Output _ -> true
         | _ -> false
 
+    member this.Copy() =
+        match this with
+        | Parameter oa -> Parameter (oa.Copy())
+        | Factor oa -> Factor (oa.Copy())
+        | Characteristic oa -> Characteristic (oa.Copy())
+        | Component oa -> Component (oa.Copy())
```
Freymaurer, line 440: What happens to Input/Output of Data?

Freymaurer (changes requested):

src/Core/Table/CompositeHeader.fs
HLWeil: Ahh the actual data entity is stored in the cells, not the header. So this is fine anyways!

HLWeil: Added some tests for data cell. Also, There already was one test which checked updating an "Input [Sample]" column to an "Input [Data]" Column.

Freymaurer (approved):

PR #478, opened 2024-11-21
Improve LDObject, add ISA model API

kMutagene:
This PR (eventually)
- slightly remodels `LDObject` to use arrays for `@type` and `additionalType`. Json encode/decode was adapted to be able to parse string OR array fields, but ALWAYS write array fields (even for singletons)
- Adds many API functions to get typed properties from ISA LDObjects (e.g., get all `Person`s as a `Person []` from the given `Investigation` LDObject)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,7 +15,7 @@
     <PackageVersion Include="FsSpreadsheet.Js"      Version="6.3.1" />
     <PackageVersion Include="FsSpreadsheet.Py"      Version="6.3.1" />
     <PackageVersion Include="YAMLicious"            Version="0.0.3" />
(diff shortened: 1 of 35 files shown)


kMutagene:
Blocked by https://github.com/CSBiology/DynamicObj/issues/35 as we need the fixed version of `CopyDynamicPropertiesTo` to correctly set only dynamic props when trying to 'upcast', e.g. a`DynamicObj` as a `LDObject`, which has some static members in contrast to a fully dynamic `DynamicObj`

HLWeil:
Rebased on main to include fix for CI Pipeline failing in linux

kMutagene:
closed in favor of #496

PR #479, opened 2024-12-03
ARC transilable async/promise based IO 

HLWeil:
- Added CrossAsync type which acts as an interface for Async in DotNET/Python and Promise in Javascript
- Added base layer of crossAsync IO helper functions
    - Python functions are put directly into the F# source code using Fables `emit` logic
    - Javascript functions are put into a js source file and called using Fables `import` logic
      - This is a future to-do, but too much of a time-eater now and the core functionality is stable enough IMO
- Added layer of crossAsync Contract handling functions
- Added top-level crossAsync ARC IO functions 
- Added top-level synchronous ARC IO functions, only for dotNET and Python as Javascript is difficult to synchronize


src/ARCtrl/ARCtrl.Python.fsproj
Freymaurer, line 1: No pypackage reference for femto?

src/ARCtrl/Xlsx.fs
```diff
@@ -1,31 +1,52 @@
-﻿namespace ARCtrl
+namespace ARCtrl
```
Freymaurer, line 1: duplicate namespace

tests/TestingUtils/TestObjects.IO/SimpleARC/.arc/.gitkeep
Freymaurer, line 1: So we keep one minimal arc in version control for testing?

Could we not write it first and therefore test correct writing too?

Then we could ommit the ARC from version control

Freymaurer: Some brainstorming: Maybe git-submodul the example arc from gitlab? So we can ensure being compatible with the example?

tests/TestingUtils/TestObjects.IO/SimpleARC/studies/experiment1_material/resources/Sample1_e36ca6b8-19ba-4504-aa82-d4781765873d.png
Freymaurer, line 1: Is this rick roll?

Freymaurer: Actually not! Now i am surprised :D

tests/TestingUtils/TestObjects.IO/SimpleARC/.arc/.gitkeep
kMutagene: > Maybe git-submodul the example arc from gitlab?

This is what we do in the integration tests for arc validation packages, see submodule [here](https://github.com/nfdi4plants/arc-validation-packages/tree/main/tests/fixtures), action that pulls submodules in CI [here ](https://github.com/nfdi4plants/actions-workflows/blob/fd66761812c42d2043758d3de0dbaa7828ca1e2b/.github/workflows/build-and-test-solution.yml#L58) and tests if submodules are there [here](https://github.com/nfdi4plants/arc-validation-packages/blob/main/tests/GitSubmodulesTests.fs)
(obviously, the test fixtures must be translated into pyxpecto, i don't know if it has that functionality)

src/ARCtrl/ARCtrl.Python.fsproj
HLWeil: We're not using femot at all. Would be a to-do for the future

tests/TestingUtils/TestObjects.IO/SimpleARC/.arc/.gitkeep
HLWeil: Yep, using the Prototype ARC would be the best solution. 

Only using the ARCs written with the library might leave a blind spot for this case.

Issue #486, opened 2025-01-21
[BUG] RO-Crate Export - ConformsTo Property


floWetzels:
In the current RO-Crate export, the `conformsTo` property is (globally) mapped to `schema.org/conformsTo`, which differs from the RO-Crate context. This leads to conflicts when flattening the jsonld object and therefore invalid RO-Crates.

HLWeil:
Can be closed? @floWetzels

floWetzels:
Fixed by PR https://github.com/nfdi4plants/ARCtrl/pull/488

Issue #485, opened 2025-01-21
[BUG] RO-Crate Export - Investigation needs DatePublished Field


floWetzels:
In the current RO-Crate export, the `datePublished` field the root data entity (i.e. the investigation of the ARC) is only written if the corresponding ISA field is set. However, it is a MUST field in the RO-Crate specification. @HLWeil and I suggest to add the timestamp of the export as a default option to always create valid RO-Crate.

HLWeil:
Can be closed? @floWetzels

floWetzels:
Fixed by PR https://github.com/nfdi4plants/ARCtrl/pull/488

Issue #484, opened 2025-01-21
[BUG] RO-Crate Export - Data files in Study not linked via hasPart


floWetzels:
If a Study contains data files, these have a corresponding `File` entry in the exported RO-Crate. However, they are only linked to via the `result` property of the `LabProcess` objects in the process sequence. They should be linked via the `hasPart` property of the `Dataset` object corresponding to the Study (as is the case for assays). Therefore, the resulting json object is not a valid RO-Crate.

HLWeil:
Can be closed? @floWetzels

floWetzels:
Fixed by PR https://github.com/nfdi4plants/ARCtrl/pull/488

Issue #473, opened 2024-11-07
[BUG] Wrong assay and study paths in RO-Crate export


floWetzels:
The identifiers of assays and studies in the json object produced by the RO-Crate export sometimes contain invalid paths. In particular, they sometimes start with a `#` symbol. This is then not valid RO-Crate.

Furthermore, the RO-Crate specification states that paths SHOULD end in a `/` symbol, which we should probably also adapt. 

**To Reproduce**
Go to the [Metabolights338_ARC](https://git.nfdi4plants.org/usadellab/metabolights338_arc) from the datahub. Export the RO-Crate in the same way the CI does (only tested with CI output).

HLWeil:
Can be closed? @floWetzels

floWetzels:
Fixed by PR https://github.com/nfdi4plants/ARCtrl/pull/488

PR #488, opened 2025-01-22
RO-Crate Fixes

floWetzels:
This PR fixes several issues with the RO-Crate export, see https://github.com/nfdi4plants/ARCtrl/issues/472, https://github.com/nfdi4plants/ARCtrl/issues/473, https://github.com/nfdi4plants/ARCtrl/issues/484, https://github.com/nfdi4plants/ARCtrl/issues/485 and https://github.com/nfdi4plants/ARCtrl/issues/486.

### DIFF
--- a/src/Json/Assay.fs
+++ b/src/Json/Assay.fs
@@ -76,7 +76,7 @@ module Assay =
             | "" -> "#EmptyAssay"
             | i -> 
                 let identifier = i.Replace(" ","_")
-                $"#assay/{identifier}"
+                $"assays/{identifier}/"
 
         let encoder (studyName:string Option) (a : ArcAssay) = 
             let fileName = Identifier.Assay.fileNameFromIdentifier a.Identifier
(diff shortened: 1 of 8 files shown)


HLWeil (approved):

HLWeil:
Looks good. As far as I can see you didn't change any fields to the RO-Crate parser should be backwards compatible?

Could you verify this? I.e. write an RO-Crate file with the last version of ARCtrl and read it in again. Alternatively try to parse an RO-Crate file from the DataHub

floWetzels:
Yes, the parser/decoder did not change. I only modified the contexts (which is irrelevant as the parser in its current state works on ARC vocabulary, not schema vocabulary), the `@id` entry of Study/Assay (which is ignored by the parser as far as I can tell) and the `dataFiles` entry of Study (which is ignored by the parser as well).

The tests check for writing and reading, right? So I think that should be fine. However, we might have to be careful and check again once ARCtrl is able to read the actual RO-Crate in its flattened and expanded form. Since this is not the case in the current version anyway, parsing should work as intended for now.

Issue #465, opened 2024-10-27
handle people address from ro-crate json 


jorgitogb:
**Describe the bug**
using JsonController.Investigation().from_rocrate_json_string(<json>) only can handle address as a str
**To Reproduce**
### A valid schema.org person:
{
    "@context": "http://schema.org",
    "@type": "Person",
    "address": {
        "@type": "PostalAddress",
        "addressCountry": "Germany",
        "addressLocality": "Musterstadt",
        "postalCode": "12345",
        "streetAddress": "Maxstraße 42"
    },
    "familyName": "Musterfrau",
    "givenName": "Erika",
    "name": "Erika Musterfrau"
}
### ARC compatible version:
{
    "@context": {
        "@vocab": "http://schema.org/"
    },
    "@type": "Person",
    "address": "Maxstraße 42, 12345 Musterstadt, Germany"
    "familyName": "Musterfrau",
    "givenName": "Erika",
}

**Expected behavior**
handle str and obj representation of an address

HLWeil:
Hey @jorgitogb,
thanks for bringing this issue up. This was also something I discussed with @zalfsten, and it's definitely something that will be handled in ARCtrl.

We're currently working on a big restructure of the whole parsing and handling of the ARC RO-Crate representation. Until that is finished (Hopefully this year), it would be nice if you could handle this on your side (e.g. put the json object as a blob into a string). 

If there's progress on this issue I will update you here.

HLWeil:
Hey @jorgitogb,

just a little update. I'm almost done with the first part of the complete overhaul of everything RO-Crate related in ARCtrl. See #496. 

This will align everything more with the [ISA-RO-Crate-Frofile](https://github.com/nfdi4plants/isa-ro-crate-profile), including also handling of these address objects.

Issue #326, opened 2024-03-19
[BUG] RO crate WIP


Freymaurer:
Sources:
- https://github.com/nfdi4plants/isa-ro-crate-profile/blob/main/profile/isa_ro_crate.md
- https://www.researchobject.org/ro-crate/profiles.html

Must still be done:

# Encoder

## Assay
- [x] Missing `creator` 
- [x] Missing `identifier` 
- [x] Missing `measurementMethod` 
- [x] Has not specified `instrument` 

#### Covered by https://github.com/nfdi4plants/isa-ro-crate-profile/issues/9?
- [ ] Has not specified `url` 

#### Covered by https://github.com/nfdi4plants/isa-ro-crate-profile/issues/5?
- [ ] Missing `headline` 
- [ ] Missing `description`
- [ ] Missing `dateCreated`
- [ ] Missing `dateModified`
- [ ] Missing `citation` 

## LabProtocol

https://github.com/nfdi4plants/isa-ro-crate-profile/blob/main/profile/isa_ro_crate.md#labprotocol

- [x] Missing `reagent`
- [x] Missing `computationalTool`
- [x] `url` has strange value "http://nfdi4plants.org/protocols/MyProtocol".
  Just a placeholder

#### Covered by https://github.com/nfdi4plants/isa-ro-crate-profile/issues/13?
- [ ] Missing `sameAs`

#### Covered by https://github.com/nfdi4plants/isa-ro-crate-profile/issues/10?
- [ ] Missing `headline`
- [ ] Has not specified `name` 

## Sample

- [x] processSequence.outputs has `additionalProperty` mapped to "characteristics" and "factorValues", instead of `or`.

![image](https://github.com/nfdi4plants/ARCtrl/assets/39732517/d856890d-6e1e-47ab-83f7-0a72e790bae7)

## Person

- [ ] requires `familiyName`, but some cultures do not have one. PLUS if we must infer a person object from a isa-json authorList string we can only set `name`. I recommend setting `familiyName` as COULD. Therefore, we should also set `givenName` as SHOULD or move the inferred name from `name` to `givenName`.

![image](https://github.com/nfdi4plants/ARCtrl/assets/39732517/64801acd-f0ca-45d3-859f-a7de0d193c51)

## ScholarlyArticle 

- [ ] `sameAs` is used for DOI in the example, but the given example doi is just a doi "11.1111/abcdef123456789", which is not of type URL.
[...]

PR #496, opened 2025-02-27
Generic LDNodes and RO-Crate parsing rework

HLWeil:
Summary
- Coming soon ™️ 

To-do:
- [ ] DOI and PubmedID as identifier
- [ ] Bioschemas context
- [ ] Some `@id` are empty (LabProtocol)
- [ ] Top-level roundabout tests
- [ ] Link datafiles instead of fragment selectors from datasets


HLWeil:
@floWetzels could you also take a look onto whether #326 is actually fixed or not (e.g. by inspecting a written RO-Crate metadata json)

floWetzels:
I tested the metadata json file generated for the ArcPrototype with the rocrate-validator. I found the following errors that need to be fixed:
- Studies und assays have wrong ids - the path to the isa-xlsx file. It should be the path to the directory ending with a `/`.
- Studies and assays are missing the `hasPart` property
- The root data entity has no `license`
- The root data entity has no `datePublished`

Furthermore,  I noticed the following when parsing the file manually:
- Paths must be relative to the root of the ro-crate
- Ad-hoc ids (e.g. for processes, paramters, parameter values) should begin with a `#` symbol
- The id of PropertyValue objects looks weird to me. It contains the `propertyID`, which could be a potential source for errors. Previously, it was generated only from `name` and `value`, if I am not mistaken.

HLWeil:
I incorporated your suggested fixes and included the test file in [d3d786e](https://github.com/nfdi4plants/ARCtrl/pull/496/commits/d3d786e6a4f8a6c5207a9e339043471e4f84d650)

kMutagene:
@HLWeil i think adding a test that performs the validation of the ro-crate file that @floWetzels did manually is needed. If not in this PR then via separate issue

src/ROCrate/Generic/File.fs
```diff
@@ -0,0 +1,123 @@
...
+        | Some (:? string as ui) -> Some ui
+        | _ -> None
+
+    static member setUsageInfoAsString(dt : LDNode, usageInfo : string, ?context : LDContext) =
+        dt.SetProperty(File.usageInfo, usageInfo, ?context = context)
+
+    static member genId(name : string) =
+        $"{name}"
+
+    static member validate(dt : LDNode, ?context : LDContext) =
```
kMutagene, line 71: Is this the equivalent of the missing file profile for https://github.com/nfdi4plants/isa-ro-crate-profile/issues/57?

src/ROCrate/Generic/Organization.fs
kMutagene, line 1: Is this the missing organization profile for https://github.com/nfdi4plants/isa-ro-crate-profile/issues/57 ?

src/ROCrate/Generic/PostalAddress.fs
kMutagene, line 1: Is this the missing postaladdress profile for https://github.com/nfdi4plants/isa-ro-crate-profile/issues/57 ? If we have 0 mandatory fields, maybe we just don't need to support this type?

src/ROCrate/Generic/PropertyValue.fs
```diff
@@ -0,0 +1,171 @@
...
+
+    static member tryGetValueReference(pv : LDNode, ?context : LDContext) =
+        match pv.TryGetPropertyAsSingleton(PropertyValue.valueReference, ?context = context) with
+        | Some (:? string as vr) -> Some vr
+        | _ -> None
+
+    static member setValueReference(pv : LDNode, valueReference : string, ?context : LDContext) =
+        pv.SetProperty(PropertyValue.valueReference, valueReference, ?context = context)
+
+    static member validate(pv : LDNode, ?context : LDContext) =
```
kMutagene, line 85: Is this missing validate functions for PropertyValue - DOI and PropertyValue - PubMedID ?

src/ROCrate/Generic/ScholarlyArticle.fs
```diff
@@ -0,0 +1,103 @@
...
+        s.SetProperty(ScholarlyArticle.identifier, identifiers)
+
+    static member getAuthors(s : LDNode, ?graph : LDGraph, ?context : LDContext) =
+        let filter ldObject context = Person.validate(ldObject, ?context = context)
+        s.GetPropertyNodes(ScholarlyArticle.author, filter = filter, ?graph = graph, ?context = context)
+
+    static member setAuthors(s : LDNode, authors : ResizeArray<LDNode>, ?context : LDContext) =
+        s.SetProperty(ScholarlyArticle.author, authors, ?context = context)
+
+    static member tryGetUrl(s : LDNode, ?context : LDContext) =
```
kMutagene, line 56: We dropped url as property of this type

src/ROCrate/ISAProfile/Assay.fs
kMutagene, line 1: Why are you deleting all these types?

tests/ROCrate/ISAProfile/Person.Tests.fs
```diff
@@ -14,7 +14,7 @@ let mandatory_properties = Person(
 let all_properties = Person(
```
kMutagene, line 14: How are these tests still passing when you deleted all the types tested? maybe not seeing the forest for the trees here

tests/ROCrate/Main.fs
```diff
@@ -3,18 +3,19 @@ module ARCtrl.ROCrate.Tests
...
-    Tests.LabProcess.main
-    Tests.LabProtocol.main
-    Tests.Sample.main
-    Tests.Data.main
-    Tests.PropertyValue.main
-    Tests.Person.main
-    Tests.ScholarlyArticle.main
+    Tests.LDContext.main
+    Tests.LDNode.main
+    //Tests.Dataset.main
```
kMutagene, line 8: hmm i think i found the issue, you are not running the tests. Why fix them then?

kMutagene (changes requested):
See comments

src/ROCrate/Generic/File.fs
HLWeil: Yes

HLWeil: ... I mean yesn't. 

It isn't missing, File = [Data](https://github.com/nfdi4plants/isa-ro-crate-profile/blob/release/profile/isa_ro_crate.md#data)

src/ROCrate/Generic/Organization.fs
HLWeil: Yes

src/ROCrate/Generic/PostalAddress.fs
HLWeil: Well there were discussions about it for the specification. Also requests for the implementation: https://github.com/nfdi4plants/ARCtrl/issues/465

I currently handle it in ARC Scaffold by just writing the json blob in the string field.

src/ROCrate/Generic/PropertyValue.fs
HLWeil: There's currently not a validate function for doi and pubmedID.

The convertes can be found here:
https://github.com/nfdi4plants/ARCtrl/blob/b755341a2b50cc7e9243c95e55ed423e6e63cfa8/src/ARCtrl/Conversion.fs#L1015

src/ROCrate/ISAProfile/Assay.fs
HLWeil: That was a misunderstanding between me, myself and I

HLWeil:
Alright the comments should be resolved. 
- I reincluded the initial ROCrate Types, and renamed the static classes using the `LD` prefix. 
- LDPropertyValue now has DOI and PubMedID helper functions
- Implemented and tested DOI and PubMedID backwards compatability (using `url` and `sameAs`)

kMutagene (approved):

PR #516, opened 2025-05-15
Add Typescript support and rework release pipeline

HLWeil:
- Javascript IO access is now handled using "emit" instead of "import", allowing removing the native helper file
- Javascript native tests now are run using vitest
- Code is now transpiled to Typescript before being packaged as Javascript with type support
- Both for python and javascript release, original pytoml and package.json are used instead of being copied somewhere
- index.ts and arctrl.py generation is now omitted in favor of static files
- PreRelease versioning is now handled separately when versioning, not when releasing. This allows for better tracking of versions and changes


Freymaurer (changes requested):
Overall very nice changes, made everything a lot cleaner

Freymaurer (approved):

PR #523, opened 2025-05-27
Python Dictionary Fixes

HLWeil:
- Improved speed of Python Dictionaries
  - RO-Crate json reading now finishes in a normal time (<1 sec) for a small ARC
    - By explicitly typing dictionary access functions 
  - Compressed json writing now finishes in a bearable time (10000 row table in <10sec)
    - By explicitly emitting native dict construction
- Reworked speed testing project
  - Now includes test preparation step for better tracking
  - Repeated testing gives better results

@Brilator @eik-dahms @Mestizia

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-﻿
+
 ![logo](logos/ARCtrl_horizontal.png)
 
 # ARCtrl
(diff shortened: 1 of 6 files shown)


HLWeil:
Maybe also to your interest, @feserm. 

As you stated, the native python dictionary is fast enough. In some cases though (where key type is generic or non-primitive), the transpiler makes use of a different, extra implementation. This ensures that everything works as expected in the transpiled code, but is pretty expensive speedwise.

In this fix we forced the usage of the native implementation where realisable. Where viable (Annotation table), we will switch from Dictionaries to Arrays, which should also be much quicker.

PR #530, opened 2025-06-04
Check for Investigation identifier presence

### LINKED ISSUE #519: [BUG] Investigation file without identifier but with any (or a made-up) key or no keys but at least 1 value results in empty ARC when using `ARC.load`
**Describe the bug**
When using `ARC.load` or (more precisely) `ArcInvestigation.fromRows` onto an ARC (or an Investigation file) where there's no Investigation identifier but either
- at least 1 key (may be correct or made-up), or
- at least 1 value

results in an empty ARC with an empty Investigation identifier.

**To Reproduce**
1. Create an ARC
2. Open the Investigation file in MS Excel or sth. alike
3. Delete everything but at least 1 key or delete everything but at least 1 value
4. Use `ARC.load` on it

**Expected behavior**
Raises error that mandatory Investigation identifier is missing.

omaus:
This PR
- changes the `fromRows` function in `ARCtrl.Spreadsheet.ArcInvestigation` module in such a way that it now checks for the presence of the Investigation identifier and fails if it is empty
  - adds the respective unit test
  - comments out unit tests `ReaderSuccessEmpty` and `WriterSuccessEmpty` since they don't make sense anymore
- fixes #519 
- has **NO** playground files added to the solution (thank God!) @HLWeil

### DIFF
--- a/src/Spreadsheet/ArcInvestigation.fs
+++ b/src/Spreadsheet/ArcInvestigation.fs
@@ -117,8 +117,10 @@ module ArcInvestigation =
             (ResizeArray remarks)
 
     let fromRows (rows : seq<SparseRow>) =
-        let en = rows.GetEnumerator()              
-        
+        if Seq.isEmpty rows then failwith "isa_investigation sheet in Investigation file is empty"
+
+        let en = rows.GetEnumerator()
+
         let emptyInvestigationInfo = InvestigationInfo.create "" "" "" "" "" []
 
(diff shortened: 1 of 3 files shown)


HLWeil (approved):
> has NO playground files added to the solution (thank God!) @HLWeil

Thank you for fulfilling this HUGE task of cleaning up the commit history. As stated in the other PR (#524), changes look good!

PR #537, opened 2025-07-09
Rework Core ArcTable towards better memory efficiency and performance

HLWeil:
These changes were necessary for much better memory efficiency, allowing Swate to depict and edit much bigger tables. 

In principle this was done in two parts:
- The ArcTable now contains a "ValueMap", which maps from the cell Hash to the actual CompositeCell. Were previously the CompositeCell was stored we now store this Hash, reducing memory usage for Cells appearing repreadetly in the same table
- Instead of a 2D canvas, we now store the Table as columns, where a column may also be represented by a constant value. For parameters, it is very often the case that we have a column with only one distinct value. In this case we just need to store it once.

Calling functions were adjusted accordingly


src/Core/Table/ArcTable.fs
```diff
@@ -15,94 +15,111 @@ type TableJoinOptions =
...
-    let mutable values = values
     member this.Headers
-        with get() = headers
-        and set(newHeaders) = 
-            SanityChecks.validate newHeaders values true |> ignore
-            headers <- newHeaders
-    /// column * row index
+        with get() = _headers
+        and set(newHeaders) =
+            // SanityChecks.validate newHeaders values true |> ignore // TODO
```
Freymaurer, line 41: Is this meant to stay "TODO"?

src/Core/Table/ArcTable.fs
```diff
@@ -372,42 +391,49 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
 
     // - Column API - //
     // GetColumnAt?
-    member this.GetColumn(columnIndex:int) : CompositeColumn =
+    member this.GetColumn(columnIndex:int, ?fillDefault : bool) : CompositeColumn =
```
Freymaurer, line 394: fillDefault behaves different to my expectations. Did you discuss this with someone else in the department and decided that we fail if a cell does not exist inside row/column count? I thought we only fail for trying to access outside these params?

src/Core/Table/ArcTable.fs
```diff
@@ -372,42 +391,49 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
...
         SanityChecks.validateColumnIndex columnIndex this.ColumnCount false
         let h = this.Headers.[columnIndex]
-        let cells = [|
-            for i = 0 to this.RowCount - 1 do 
-                match this.TryGetCellAt(columnIndex, i) with
-                | None -> failwithf "Unable to find cell for index: (%i, %i)" columnIndex i
-                | Some c -> c
-        |]
+        let cells =
+            [|
```
Freymaurer, line 399: This can be optimized by first checking if columnIndex points to a `Constant` columnRef, then we can just fill with the same value

src/Core/Table/ArcTable.fs
```diff
@@ -468,105 +494,107 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
             newTable
 
     // - Row API - //
-    member this.AddRow (?cells: CompositeCell [], ?index: int) : unit = 
+    member this.AddRow (?cells: ResizeArray<CompositeCell>, ?index: int) : unit =
         let index = defaultArg index this.RowCount
-        let cells = 
+        let cells =
```
Freymaurer, line 499: Again this behaves different then my expecatation. Did we not decide on a sparse approach?

Freymaurer: Will wait for your answers before continuing the review.

src/Core/OntologyAnnotation.fs
```diff
@@ -19,17 +19,30 @@ type OntologyAnnotation(?name,?tsr,?tan, ?comments) =
         | tan -> tan
     let mutable _comments : ResizeArray<Comment> = defaultArg comments <| ResizeArray()
 
+    let mutable _tanInfo : {| IDSpace : string; LocalID : string |} option =
```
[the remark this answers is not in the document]
HLWeil: It's not my preferred solution either, but the TanInfo is relatively expensive to compute but quite important as a basis for the HashCode. 
By allowing this info to be set from the outside we can omit multiple computations of this value in some cases.

src/Core/Table/ArcTable.fs
HLWeil: I mean, with the `fillDefault` flag every user can decide. I could add some comment to make it more clear. Or would you wish to reverse the default behavious, and have a `failIfMissing` flag instead?

src/Core/Table/ArcTable.fs
HLWeil: True that, this should just increase the rowCount of the "Values"

Freymaurer (changes requested):

src/Core/Table/ArcTableAux.fs
```diff
@@ -251,112 +613,112 @@ module Unchecked =
...
-        let columnCount = getColumnCount headers
-        let previousRowCount = getRowCount values
-        // iterate over columns
-        for columnIndex = 0 to columnCount - 1 do
-            let lastValue = values[columnIndex,previousRowCount-1]
-            for rowIndex = previousRowCount - 1 to rowCount - 1 do
-                setCellAt (columnIndex,rowIndex,lastValue) values
-
-    let addRow (index:int) (newCells:CompositeCell []) (headers: ResizeArray<CompositeHeader>) (values:Dictionary<int*int,CompositeCell>) =
+    let fillMissingCells (headers: ResizeArray<CompositeHeader>) (values:ArcTableValues) =
```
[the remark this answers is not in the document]
HLWeil: I still kept the function, it will fill in default cells like before. Might be interesting in cases where we expect to make many random access calls and want to evaluate default values beforehand. Not used by any of our own functions.

src/Core/Table/ArcTable.fs
```diff
@@ -662,29 +690,29 @@ type ArcTable(name: string, headers: ResizeArray<CompositeHeader>, values: Syste
             copy
 
     ///
-    member this.AddProtocolTypeColumn(?types : OntologyAnnotation [], ?index : int) =
+    member this.AddProtocolTypeColumn(?types : ResizeArray<OntologyAnnotation>, ?index : int) =
```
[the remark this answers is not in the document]
HLWeil: Added forceReplace flag

PR #553, opened 2025-10-08
Add license logic to `ARC` class + tests :sparkles:

### LINKED ISSUE #552: [Feature Request] Add License field to ARC type and add contract handling
We require settable options for the license field in RO crate. Currently it is only given as a default.

I propose adding an optional field to the ARC type which contains the full LICENSE file content as string. I will add comments and tests for this scenario and try to create it so that we can support different formats in the near future.

Freymaurer:
Closes #552 

## Added

- ✨ License mutable field to ARC class
- ✨ License class with some future proves, by already defining type of License class (for now only `LicenseContentType.Fulltext`
- ✨ Updated RO-Crate JSON logic, as well as scaffold IO logic to use new license field.
- ✅ Tests for RO-Crate and Contract creation logic

## Changed

- ✅  Updated IO tests in `tests/ARCtrl` to empty the target folders before writing into them. This was previously only done by the full build chain (to my knowledge). So running `dotnet watch run` inside the folder was only successful the first time, then errored because files already existed

### DIFF
--- a/src/ARCtrl/ARC.fs
+++ b/src/ARCtrl/ARC.fs
@@ -664,7 +664,7 @@ type ARC(identifier : string, ?title : string, ?description : string, ?submissio
         // Map containing the DTOTypes and objects for the ISA objects.
         if this.StaticHash = 0 then
             this.StaticHash <- this.GetLightHashCode()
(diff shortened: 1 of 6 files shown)


HLWeil (commented):
Put License Text into `text` field of `CreativeWork` object. id of object should be name of LICENSE file

PR #562, opened 2025-10-28
Add CWL encoding

caroott:
Add functions to encode the ARCtrl CWLType back to YAML


src/CWL/Encode.fs
```diff
@@ -1 +1,327 @@
...
+/// Each decode path gets a corresponding encode path that produces
+/// YAML key ordering and scalar formatting matching existing fixtures.
+module Encode =
+    // ------------------------------
+    // Generic YAML node constructors
+    // ------------------------------
+    let private yScalar (s:string) =
+        YAMLElement.Value { Value = s; Comment = None }
+
+    let private yInt (i:int) = YAMLElement.Value { Value = string i; Comment = None }
```
HLWeil: Shouldn't these generic functions be part of YAMLicious?

caroott: done

PR #570, opened 2025-11-14
CWL Integration

HLWeil:
First full implementation of CWL integration into the ARCtrl datamodel and conversion to RO-Crate datamodel according to [arc-wr-ro-crate-profile](https://github.com/nfdi4plants/arc-wr-ro-crate-profile).

### Changes to Core Datamodel
- ArcWorkflow type now has 
  - `cwlDescription` field which contains parsed content of `workflow.cwl`
  - `parameters` field containing `ResizeArray<OntologyAnnotation>` (Important for SWATE, @Freymaurer)
- ArcRun now has 
  - `cwlDescription` field which contains parsed content of `run.cwl`
  - `cwlInput` field which contains parssed content of `run.yml`
- Scaffold parser and contract handling was extended and adjusted accordingly, tests included

### Changes to RO-Crate Datamodel
- Added Static Classes for `Schema.org`/`Bioschemas` types:
  - `CreateAction`
  - `ComputationalWorkflow`
  - `FormalParameter`
  - `SoftwareSourceCode`
- Added Static Classes for Combi types:
  - [WorkflowProtocol](https://github.com/nfdi4plants/arc-wr-ro-crate-profile/blob/release/profile/arc_wr_ro_crate.md#workflow-protocol)
  - [WorkflowInvocation](https://github.com/nfdi4plants/arc-wr-ro-crate-profile/blob/release/profile/arc_wr_ro_crate.md#workflow-invocation)
- Several adjustments to existing classes to accomodate new classes and converters

### RO-Crate conversion
- Added two new static types:
  - `WorkflowConversion` containing roundabout conversion functions for 
    - CWL Inputs and Outputs to FormalParameters 
    - ISA Components to ComputationTools
    - CWL ProcessingUnit (ToolDescription or Workflow) to WorkflowProtocol
    - ARC Workflow to Dataset
  - `RunConversion` containing roundabout conversion functions for 
    - YML Input Values to PropertyValues
    - CWL Workflow and YML Input Values to WorkflowInvocation
    - ARC Run to Dataset
- Things to consider
  - No reliable linking between converted workflow entities and conversion is only one layer deep
  - RO-Crate model is just a subset of the CWL datamodel, roundabout will therefore lose information on non-basic CWL types. Do we need a mechanism to retreive the information from the file again? Override should be prevented
  - Still not sure about conversion of WorkflowSteps and Requirements
  - Both the prospective CWL annotation of workflows and their given input values (run.yml) and the retrospective annotation of Run Annotation tables are parsed to `WorkflowInvocation` nodes. Currently, they are put into the same list and treated equally in principle. To distinguish between them and parse them back correctly, I rely on some marker fields. I.e. the existence of `exampleOfWork` on those object nodes which point to CWL Input parameters
  - The `Wrapper CWL` logic is NOT yet included and will be content of a future PR

### Other Small Changes
- Updated DynamicObj dependency to 7.1.0
- Made some CWL Record types to Classes and added AttachMembers Attribute
- Added some custom equality function to CWL classes
- Tests, but not extensive

### DIFF
--- a/src/ARCtrl/Conversion/Workflow.fs
+++ b/src/ARCtrl/Conversion/Workflow.fs
@@ -178,6 +178,7 @@ type WorkflowConversion =
         LDWorkflowProtocol.create(
             id = filePath,
             ?inputs = inputs,
+            programmingLanguages = ResizeArray.singleton (LDComputerLanguage.createCWL()),
             outputs = outputs,
             additionalType = ResizeArray [WorkflowConversion.toolDescriptionTypeName]
         )
@@ -214,6 +215,7 @@ type WorkflowConversion =
             id = filePath,
             inputs = inputs,
             outputs = outputs,
+            programmingLanguages = ResizeArray.singleton (LDComputerLanguage.createCWL()),
             hasParts = steps,
             additionalType = ResizeArray [WorkflowConversion.workflowDescriptionTypeName]
         )
--- a/src/ROCrate/ARCtrl.ROCrate.fsproj
+++ b/src/ROCrate/ARCtrl.ROCrate.fsproj
@@ -18,6 +18,7 @@
     <Compile Include="LDTypes\Comment.fs" />
     <Compile Include="LDTypes\DefinedTerm.fs" />
     <Compile Include="LDTypes\FormalParameter.fs" />
+    <Compile Include="LDTypes\ComputerLanguage.fs" />
     <Compile Include="LDTypes\PropertyValue.fs" />
     <Compile Include="LDTypes\PostalAddress.fs" />
     <Compile Include="LDTypes\Organization.fs" />
--- a/src/ROCrate/LDTypes/ComputerLanguage.fs
+++ b/src/ROCrate/LDTypes/ComputerLanguage.fs
@@ -1,0 +1,146 @@
+namespace ARCtrl.ROCrate
+
+open DynamicObj
+open Fable.Core
+open ARCtrl.ROCrate
+
+[<AttachMembers>]
+type LDComputerLanguage =
+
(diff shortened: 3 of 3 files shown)


src/ROCrate/Generic/WorkflowInvocation.fs
kMutagene: i don't understand why this is in `Generic`, while there is `Workflow` in `ARCRun` in `WorkflowProfile`, but not `ARCWorkflow`. The folder name `WorkflowProfile` is also confusing. Do you mean our `ARC Workflow Run` profile, or the bioschemas `Workflow` profile?

kMutagene: I think in general, a readme in this `ROCrate` project/folder structure about what is found where would be very helpful.

HLWeil: This seems like a copy error. The "WorkflowProfile" folder is unused. I agreethe folder name "Generic" is a bit weird. Maybe I could just rename it to "SchemaTypes"?

HLWeil:
> > One important thing to note. The current Workflow and Run parser will only create the according objects, if respectively an isa.run.xlsx or isa.workflow.xlsx file are present.
> > The RO-Crate writer will fail, if no workflow.cwl or run.cwl were given.
> 
> Confusing for me. Does this mean no ro crate export is possible if there are no workflows/runs? That would be bad

If not run.xlsx or workflow.xlsx is present in a run or workflow folder, the given worklfow or run object will not be included in the RO-Crate output.
If the xlsx file exists, but no matching run.cwl or workflow.cwl exists, the export to RO-Crate will fail. Open for discussion...

PR #606, opened 2026-03-30
bump to version 3.0.4

caroott:


HLWeil (approved):
I reran the pipeline. The fail was just a coincidence with an unreachable online ressource

PR #621, opened 2026-06-21
 GetWriteContracts fix

HLWeil:
Rework ARC.GetWriteContracts behaviour, so it does not lead to overwriting payload files


HLWeil:
Not sure about READMEs... The initial idea was to create them together with the ARC, but we don't want to overwrite it with an empty string. Should we skip writing this for now? 
@Freymaurer @Etschbeijer

Freymaurer:
I think the best solution is to skip write contracts if they already exist and only run ipdate contracts on existing files. This allows us to create files with default content, as well as ensure they exist without overwriting anything we do not specifically want to overwrite

PR #628, opened 2026-07-29
Update to Fable 5 / .NET 10

caroott:
Bumps Fable 4.24 → 5.13, .NET SDK 8 → 10, FSharp.Core 10, and everything that had to move with it (FsSpreadsheet 7.0.1, Thoth, DynamicObj 8, YAMLicious 1.0.0, Pyxpecto 2, fable-library 2.5.1 on npm).

Python string hashing was broken. Fable's string_hash saturates at Int32.MaxValue for strings of ~10 chars or more, so any two longer strings hashed the same and distinct OntologyAnnotations compared equal. Added a djb2 hashString with explicit 32-bit wrapping that gives the same result on .NET, JS and Python.

Also had to add Py.Stringable/Comparable/Equatable/Hashable markers in a few places, otherwise string x prints the default object repr and comparing arrays of OntologyAnnotation falls through to Python's own ordering. There's a comment at each one saying why, so they can go once Fable fixes it upstream.

The TS build now uses fable-library-ts instead of -js. Fable emits imports with .ts extensions and the js package only ships .js, so nothing resolved.

CWL parameter references used to hold a flat ResizeArray<string> and lost all type info. There's a real CWLParameterValue union now (Null/String/Int/Float/Boolean/File/Directory/Array/Record) plus encoding for it. Values still exists and flattens to/from strings so nothing downstream breaks.

Renames to watch out for: Value → PropertyValue (name collided under Fable 5), and DTO.Spreadsheet is now FsWorkbook instead of obj.


src/Core/PropertyValue.fs
```diff
@@ -8,15 +8,15 @@ open ARCtrl.Helper
 
 
 [<AttachMembers>]
-type Value =
+type PropertyValue =
```
HLWeil: The rename gives the impression, that this type is equivalent to https://schema.org/PropertyValue, which it is not. I would suggest renaming it to something along the lines of `BasicValue`, `ScalarValue` or `AtomicValue`

src/Contract/Contract.fs
```diff
@@ -36,7 +37,7 @@ type CLITool =
 #endif
 [<RequireQualifiedAccess>]
 type DTO =
-    | Spreadsheet of obj
+    | Spreadsheet of FsWorkbook
```
HLWeil, line 40: good!

src/ARCtrl/Conversion/Person.fs
```diff
@@ -51,7 +51,7 @@ type PersonConversion =
     static member composePerson (person : ARCtrl.Person, ?persons : ARCtrl.Person seq) =
         let orcid =
             match person.ORCID with
-            | Some o ->
```
HLWeil, line 54: Very important!

HLWeil (changes requested):
First round of review

src/ARCtrl/Conversion/Person.fs
caroott: Yeah, i was lucky to catch that in time

src/Core/PropertyValue.fs
caroott: [f97ec91](https://github.com/nfdi4plants/ARCtrl/pull/628/commits/f97ec91af8c3af738285fb4ac85d6029486330ff)

the filename already gave that impression before, but i renamed it

src/ARCtrl/Conversion/Run.fs
```diff
@@ -27,6 +66,69 @@ type RunConversion =
...
+    static member tryGetNonNullUnionType (type_ : CWL.CWLType) =
+        match type_ with
+        | CWL.CWLType.Union types ->
+            types
+            |> Seq.tryFind (function
+                | CWL.CWLType.Null -> false
+                | _ -> true)
+        | _ -> Some type_
+
+    static member private tryParseInt64 (value: obj) =
```
HLWeil: Do these functions really belong here?

src/ARCtrl/Conversion/Run.fs
```diff
@@ -14,6 +15,44 @@ open ARCtrl.Helper.Regex.ActivePatterns
 
 type RunConversion = 
 
+    static member private tryDynamicString name (value: DynamicObj) =
```
HLWeil: Could you explain the changes here? Seems like a lot of boilerplate that would rather belong in the CWL repo and a lot just for correct primitive type handling

src/ARCtrl/ARC.fs
```diff
@@ -129,6 +129,18 @@ module ARCAux =
             |> FileSystem.create
         fs.Union(tree)
 
+    /// Backing value for ARC.DefaultContracts.
```
HLWeil, line 132: Is this a hotfix for a Fable bug?

src/Core/Table/CompositeHeader.fs
```diff
@@ -23,9 +23,12 @@ type IOType =
         Material
     |]
 
-    static member Cases = 
-        Microsoft.FSharp.Reflection.FSharpType.GetUnionCases(typeof<IOType>) 
-        |> Array.map (fun x -> x.Tag, x.Name)
+    // Written without `|>`: Fable's Python backend fails to transform a piped static property
```
HLWeil, line 26: In this case, one could argue it's just style, but still also kind of a hotfix against FABLE? Is there and issue, do we expect this to be fixed?

src/Core/ARCtrl.Core.fsproj
```diff
@@ -21,7 +21,7 @@
     <Compile Include="Comment.fs" />
     <Compile Include="CommentList.fs" />
     <Compile Include="OntologyAnnotation.fs" />
-    <Compile Include="Value.fs" />
+    <Compile Include="PropertyValue.fs" />
```
HLWeil: Filename still PropertyValue

src/Package.Metadata.props
```diff
@@ -16,6 +16,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="FSharp.Core" />
```
HLWeil, line 19: Why, I don't think we want explicit FSharp.Core references anywhere

HLWeil (changes requested):

src/ARCtrl/ARC.fs
caroott: Fable bug. See comment below

src/Core/Table/CompositeHeader.fs
caroott: yes, fixed for fable, but if we rewrite it there is no detriment for us as far as i can see. we would need to open an issue with them for a fix. same is true for other fable fixes.

src/Core/ARCtrl.Core.fsproj
caroott: [52aa418](https://github.com/nfdi4plants/ARCtrl/pull/628/commits/52aa41819a667b28258269357d31aa77d14e042e)

src/Package.Metadata.props
caroott: central package management disables the SDK's implicit FSharp.Core reference, so without this we fall back to Fable.Core's transitive >= 4.7.2 and the build fails on string interpolation. It's versionless, the version still comes from Directory.Packages.props

src/ARCtrl/Conversion/Run.fs
caroott: DynamicObject had backingfield problems, which were fixed on main while this PR was still in work. It can be removed [7e07479](https://github.com/nfdi4plants/ARCtrl/pull/628/commits/7e07479ad848b050a8951d5ae302e40792e72e1a)

src/ARCtrl/Conversion/Run.fs
caroott: moved: [e1c9ca0](https://github.com/nfdi4plants/ARCtrl/pull/628/commits/e1c9ca0a09ffe31debb8efb3b90aa6c734b098ca)

src/Core/PropertyValue.fs
HLWeil: Oh I see, sry and thanks!

HLWeil (approved):
Thanks for the efforts

